PR #6, opened 2021-03-19
Create skeleton code for spoke agent

elgnay:
Signed-off-by: Yang Le <yangle@redhat.com>


go.mod
```diff
@@ -0,0 +1,17 @@
...
+go 1.15
+
+require (
+	github.com/go-bindata/go-bindata v3.1.2+incompatible
+	github.com/open-cluster-management/api v0.0.0-20201210143210-581cab55c797
+	github.com/openshift/build-machinery-go v0.0.0-20200917070002-f171684f77ab
+	github.com/openshift/library-go v0.0.0-20201207213115-a0cd28f38065
+	github.com/spf13/cobra v1.1.1
+	github.com/spf13/pflag v1.0.5
+	k8s.io/apimachinery v0.19.5
```
qiujian16: let's move to 0.20 i think

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM docker.io/openshift/origin-release:golang-1.14 AS builder
+WORKDIR /go/src/github.com/open-cluster-management/addon-framework
+COPY . .
+ENV GO_PACKAGE github.com/open-cluster-management/addon-framework
+
+RUN make build --warn-undefined-variables
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+COPY --from=builder /go/src/github.com/open-cluster-management/addon-framework/addon-manager /
+RUN microdnf update && microdnf clean all
```
qiujian16: eol

pkg/spoke/controllers/lease/lease_controller.go
```diff
@@ -0,0 +1,73 @@
+package lease
```
qiujian16: I do not think we need this for the current design?

go.mod
elgnay: Use 0.20.x instead

Dockerfile
elgnay: Added.

pkg/spoke/controllers/lease/lease_controller.go
elgnay: Removed.

PR #7, opened 2021-03-25
Setup framework interface

qiujian16:


pkg/agent/inteface.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// GetAgentAddonOptions returns the agent options.
+	GetAgentAddonOptions() AgentAddonOptions
+}
+
+// AgentAddonWithRegistration manifests of agent deployed on managed cluster, rbac rules and bootstrap secret for agent to register to hub.
+type AgentAddonWithRegistration interface {
+	AgentAddon
+	// HubRBAC returns the role and rolebinding on the hub for an agent on the managed cluster.
+	HubRBAC(cluster *clusterv1.ManagedCluster, group string) (*rbacv1.Role, *rbacv1.RoleBinding)
```
elgnay: Do we need to support clusterrole/clusterrolebindnig?

qiujian16: added

cmd/helloworld/main.go
```diff
@@ -0,0 +1,127 @@
...
+func (h *helloWorldAgent) GetRegistrationOption() *agent.RegistrationOption {
+	// return nil since registration is not needed by this agent
+	return nil
+}
+
+// Another agent with registration enabled.
+type helloWorldAgentWithRegistration struct{}
+
+func (h *helloWorldAgentWithRegistration) Manifests(cluster *clusterv1.ManagedCluster) ([]runtime.Object, error) {
+	return []runtime.Object{
```
deads2k: I think you may want to show how to use yaml files read in as well.  That gives people the option of describing it both ways.

qiujian16: yes, I added an example to use  yaml  files.

PR #9, opened 2021-04-09
Addon impl

qiujian16:


pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -0,0 +1,230 @@
...
+			Status:  metav1.ConditionFalse,
+			Reason:  "ManifestWorkApplyFailed",
+			Message: fmt.Sprintf("failed to apply manifestwork: %v", err),
+		})
+		c.addonClient.AddonV1alpha1().ManagedClusterAddOns(managedClusterAddonCopy.Namespace).UpdateStatus(
+			ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
+		return err
+	}
+
+	// Update addon status based on work's status
```
elgnay, line 135: A condition with type `Progressing` is appended to ManagedClusterAddOn by klusterlet-addon-controller. See [here](https://github.com/open-cluster-management/klusterlet-addon-controller/blob/a836e6f94d19156e460f1433aa2df34a313b09ba/pkg/controller/managedclusteraddon/managed_cluster_addon_controller.go#L257) for more details. Should we support this condition type?

elgnay:
Looks pretty good! 

One thing I want to mention is the default value of users/groups. In registration repo, we have different default value used.
```go
func defaultCommonName(clusterName, agentName, addonName string) string {
	return fmt.Sprintf("%s:agent:%s", defaultOrganization(clusterName, addonName), agentName)
}

func defaultOrganization(clusterName, addonName string) string {
	return fmt.Sprintf("system:open-cluster-management:cluster:%s:addon:%s", clusterName, addonName)
}
```
Should we make registration repo align with the default configuration of addon-framework?

pkg/addonmanager/controllers/agentdeploy/controller.go
qiujian16: it is a bit different, since it is not necessary that a manifestwork is created for the addon. It is possible that addon does not have any manifestwork.

examples/helloworld/helloworldagent.go
```diff
@@ -0,0 +1,154 @@
...
+func newAgentController(
+	spokeKubeClient kubernetes.Interface,
+	configmapInformers corev1informers.ConfigMapInformer,
+	clusterName string,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &agentController{
+		spokeKubeClient:    spokeKubeClient,
+		clusterName:        clusterName,
+		hunConfigMapLister: configmapInformers.Lister(),
```
porridge, line 115: @qiujian16 what does `hun` stand for here? Should this have been `hub` perhaps? 🤔

PR #21, opened 2021-06-09
update readme

elgnay:
Signed-off-by: Yang Le <yangle@redhat.com>


qiujian16 (commented):
/lgtm

I would think we also need some document on
What is this?
How a developer would be able develop an addon using this. 

It could be in a docs directory, and a link in readme

PR #23, opened 2021-06-15
Lease compatible

qiujian16:


pkg/lease/lease_controller.go
```diff
@@ -52,14 +55,21 @@ func (r *leaseUpdater) Start(ctx context.Context) {
...
+	if err != nil {
+		return err
 	}
-	lease, err := r.kubeClient.CoordinationV1().Leases(r.leaseNamespace).Get(context.TODO(), r.leaseName, metav1.GetOptions{})
+
+	return r.updateLease(ctx, hubClient)
+}
+
+func (r *leaseUpdater) updateLease(ctx context.Context, client kubernetes.Interface) error {
+	lease, err := r.kubeClient.CoordinationV1().Leases(r.leaseNamespace).Get(ctx, r.leaseName, metav1.GetOptions{})
```
elgnay: What's the value of `leaseNamespae` here? On hub it should be cluster namespace, while on managed cluster, it should be the installation namespace of addon.

qiujian16: that is right.

pkg/lease/lease_controller.go
```diff
@@ -75,21 +85,51 @@ func (r *leaseUpdater) reconcile(ctx context.Context) {
...
+	version, err := r.kubeClient.Discovery().ServerVersion()
+	if err != nil {
+		klog.Errorf("Failed to get kubernetes version: %v", err)
+	}
+	kubeVersion, err := versionutil.ParseGeneric(version.String())
+	if err != nil {
+		klog.Errorf("Failed to parse kubernetes version: %v", err)
+	}
+
+	if cnt, err := kubeVersion.Compare("v1.12.0"); err == nil && cnt < 0 {
```
elgnay: v1.12.0 or v1.17.0?

qiujian16: updated, i think we may still update local lease at first, and if it fails, fallback to hub lease.

pkg/lease/lease_controller.go
```diff
@@ -75,21 +83,42 @@ func (r *leaseUpdater) reconcile(ctx context.Context) {
 				},
 			},
 		}
-		if _, err := r.kubeClient.CoordinationV1().Leases(r.leaseNamespace).Create(context.TODO(), lease, metav1.CreateOptions{}); err != nil {
-			klog.Errorf("unable to create addon lease %q/%q on hub cluster. error:%v", r.leaseNamespace, r.leaseName, err)
+		if _, err := client.CoordinationV1().Leases(r.leaseNamespace).Create(ctx, lease, metav1.CreateOptions{}); err != nil {
```
elgnay: Should we use r.leaseNamespace or namespace here?

pkg/lease/lease_controller.go
```diff
@@ -75,21 +83,42 @@ func (r *leaseUpdater) reconcile(ctx context.Context) {
...
 	case err != nil:
-		klog.Errorf("unable to get addon lease %q/%q on hub cluster. error:%v", r.leaseNamespace, r.leaseName, err)
-		return
+		return err
 	default:
 		//update lease
 		lease.Spec.RenewTime = &metav1.MicroTime{Time: time.Now()}
-		if _, err = r.kubeClient.CoordinationV1().Leases(r.leaseNamespace).Update(context.TODO(), lease, metav1.UpdateOptions{}); err != nil {
-			klog.Errorf("unable to update cluster lease %q/%q on hub cluster. error:%v", r.leaseNamespace, r.leaseName, err)
+		if _, err = client.CoordinationV1().Leases(r.leaseNamespace).Update(context.TODO(), lease, metav1.UpdateOptions{}); err != nil {
```
elgnay: r.leaseNamespace or namespace?

pkg/lease/lease_controller.go
qiujian16: updated

pkg/lease/lease_controller.go
qiujian16: updated

PR #32, opened 2021-08-02
docs(readme): correct image name for demo addon

zhujian7:
image name should be `quay.io/open-cluster-management/helloworld-addon:latest`
instead of `quay.io/open-cluster-management-io/helloworld-addon:latest`,
since the default value of `IMAGE_REGISTRY` in [Makefile](https://github.com/open-cluster-management-io/addon-framework/blob/main/Makefile#L27) is `quay.io/open-cluster-management`


zhujian7:
cc @elgnay @qiujian16

qiujian16:
Thanks!

PR #34, opened 2021-09-08
add config checker

xuezhaojun:

### DIFF
--- a/pkg/utils/config_checker.go
+++ b/pkg/utils/config_checker.go
@@ -1,0 +1,94 @@
+package utils
+
+import (
+	"crypto/md5"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"sync"
+
+	"k8s.io/apiserver/pkg/server/healthz"
+)
+
+var _ healthz.HealthChecker = &configChecker{}
+
+// configChecker is used to notify container to restart when config files updated
+type configChecker struct {
+	name        string
+	configfiles []string
+	checksum    [16]byte
+	sync.Mutex
+}
+
+// NewConfigChecker
+// name could be any string.
+// configfiles should be the same as your target container are using now.
+//
+// Example:
+// type configCheckerServer struct {
+// 	checkers []heathz.HealthChecker
+// }
+//
+// func NewConfigCheckerServer(checkers []healthz.HealthChecker) *configCheckerServer {
+// 	return &configCheckerServer{checkers: checkers}
+// }
+//
+// func (s *configCheckerServer) ServerHttp(rw http.ResponseWriter, r *http.Request) {
+// 	for _, c := range s.chekers {
+// 		if c.Name() == r.URL {
+// 			if err := c.Check(); err != nil {
+// 				rw.WriteHeader(500)
+// 			} else {
+// 				rw.WriteHeader(200)
+// 			}
+// 		}
+// 	}
+// }
+//
+// Containers in the same pod of configCheckServer can add liveness probe and be triggered restart as config files be modified.
+func NewConfigChecker(name string, configfiles ...string) (*configChecker, error) {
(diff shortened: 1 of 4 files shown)


xuezhaojun:
This is a little demo how we leverage config watcher:
```go
const configmapfile = "/tmp/configmap/key"
const secretfile = "/tmp/secret/key"

func main() {
         // -- this part should add in the container need restart when config change -- 
	watcher, err := utils.NewConfigWatcher(8080, "/unchange", configmapfile)
	if err != nil {
		log.Fatal(err)
	}
	go watcher.Serve()
         // ---------------------------------------------------------------------------

	configMapContent, err := ioutil.ReadFile(configmapfile)
	if err != nil {
		log.Fatal(fmt.Errorf("read configmap failed:%v", err))
	}

	secretContent, err := ioutil.ReadFile(secretfile)
	if err != nil {
		log.Fatal(fmt.Errorf("read configmap failed:%v", err))
	}

	http.HandleFunc("/hello", func(rw http.ResponseWriter, r *http.Request) {
		rw.WriteHeader(200)
		rw.Write(configMapContent)
		rw.Write(secretContent)
	})

	http.ListenAndServe(":8081", nil)
}
```

If we create a deployment like the following:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - name: config-watcher
        image: config-watcher:1.0.5
        command: ["./config-watcher"]
        volumeMounts:
          - name: test-secret
            mountPath: /tmp/secret
            readOnly: true
          - name: test-configmap
            mountPath: /tmp/configmap
            readOnly: true
        livenessProbe:
          httpGet:
            scheme: HTTP
            port: 8080
            path: /unchange
          initialDelaySeconds: 30
          timeoutSeconds: 30
      volumes:
        - name: test-secret
          secret:
            secretName: test
        - name: test-configmap
          configMap:
            name: test
```

With this, the container in pod could restart itself after 30 seconds when configmap or secret is update

xuezhaojun:
The usage of config checker is like:
```go
type configCheckerServer struct {
  checkers []heathz.HealthChecker
}

func NewConfigCheckerServer(checkers []healthz.HealthChecker) *configCheckerServer {
  return &configCheckerServer{ checkers: checkers}
}

func (s *configCheckerServer) ServerHttp(rw http.ResponseWriter, r *http.Request) {
  for _, c := range s.chekers {
    if c.Name() == r.URL {
      if err := c.Check(); err != nil { rw.WriteHeader(500) } else { rw.WriteHeader(200) }
    }
  }
}
```

anp-server liveness probe access `/anp-server` and user-server livenss probe access `/user-server`

But this also means there will be an inconsistent error if anyone else accesses anp config checker.

xuezhaojun:
@qiujian16 Comments have been added, please have a view

pkg/utils/config_checker.go
```diff
@@ -0,0 +1,90 @@
...
+// Name return the name fo the configChecker
+func (c *configChecker) Name() string {
+	return c.name
+}
+
+// Check would return nil if current configfiles's checksum is equal to cached checksum
+// If checksum not equal, it will return err and update cached checksum with current checksum
+// Note that: configChecker performs a instant update after it returns err, so DO NOT use one configChecker for multible containers!!!
+func (cc *configChecker) Check(_ *http.Request) error {
+	newChecksum, err := load(cc.configfiles)
```
qiujian16, line 70: I think you might want to add a lock around this to ensure it is thread safe

xuezhaojun: Added locker.

qiujian16 (commented):
/approve
/lgtm

PR #36, opened 2021-09-16
Add an interface to deploy addon

qiujian16:


pkg/addonmanager/controllers/addoninstall/controller.go
```diff
@@ -0,0 +1,126 @@
...
+		},
+		func(obj interface{}) bool {
+			accessor, _ := meta.Accessor(obj)
+			if _, ok := c.agentAddons[accessor.GetName()]; !ok {
+				return false
+			}
+
+			return true
+		},
+		addonInformers.Informer()).
```
elgnay, line 60: The controller may not need to watch the changes on ManagedClusterAddons.

**Update:** It monitors changes of ManagedClusterAddons in case the CR is modified. So `WithFilteredEventsInformersQueueKeyFunc` should return the addon namespace (cluster name) instead of `cache.MetaNamespaceKeyFunc(obj)`.

pkg/addonmanager/controllers/addoninstall/controller.go
```diff
@@ -0,0 +1,126 @@
...
+		WithInformersQueueKeyFunc(
+			func(obj runtime.Object) string {
+				accessor, _ := meta.Accessor(obj)
+				return accessor.GetName()
+			},
+			clusterInformers.Informer(),
+		).
+		WithSync(c.sync).ToController(fmt.Sprintf("addon-deploy-controller"), recorder)
```
elgnay: "addon-install-controller"?

PR #40, opened 2021-10-13
add reload mode

xuezhaojun:
add parameter 'reload' to specify whether we should reload if non-equal occur

### DIFF
--- a/pkg/utils/config_checker.go
+++ b/pkg/utils/config_checker.go
@@ -17,6 +17,7 @@ type configChecker struct {
 	name        string
 	configfiles []string
 	checksum    [16]byte
+	reload      bool
 	sync.Mutex
 }
 
@@ -55,7 +56,15 @@ func NewConfigChecker(name string, configfiles ...string) (*configChecker, error
 		name:        name,
 		configfiles: configfiles,
 		checksum:    checksum,
+		reload:      false,
 	}, nil
+}
+
+// SetReload can update the ‘reload’ fields of config checker
+// If reload equals to false, config checker won't update the checksum value in the cache, and function Check would return error forever if config files are modified.
+// but if reload equals to true, config checker only returns err once, and it updates the cache with the latest checksum of config files.
+func (c *configChecker) SetReload(reload bool) {
+	c.reload = reload
 }
 
 // Name return the name fo the configChecker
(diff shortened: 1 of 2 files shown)


pkg/utils/config_checker.go
```diff
@@ -55,9 +56,14 @@ func NewConfigChecker(name string, configfiles ...string) (*configChecker, error
 		name:        name,
 		configfiles: configfiles,
 		checksum:    checksum,
+		reload:      true,
```
qiujian16: I think you would want to set it as false by default

pkg/utils/config_checker.go
```diff
@@ -55,9 +56,14 @@ func NewConfigChecker(name string, configfiles ...string) (*configChecker, error
...
 	}, nil
 }
 
+func (c *configChecker) Reload(reload bool) {
```
qiujian16: change name to SetReload or WithReload, and add a comment on what it really means.

pkg/utils/config_checker.go
xuezhaojun: Sure we can, but it would require us to add SetReload() in cluster-proxy-addon since previously we used configchecker with reload default.

pkg/utils/config_checker.go
xuezhaojun: Sure. It should come with more comments.

pkg/utils/config_checker.go
qiujian16: only if you update the deps.

xuezhaojun: Sure, we can do that.

pkg/utils/config_checker.go
```diff
@@ -55,9 +56,17 @@ func NewConfigChecker(name string, configfiles ...string) (*configChecker, error
 		name:        name,
 		configfiles: configfiles,
 		checksum:    checksum,
+		reload:      false,
 	}, nil
 }
 
+// SetReload can update the reload filed of config checker
```
qiujian16: type filed

pkg/utils/config_checker.go
```diff
@@ -55,9 +56,17 @@ func NewConfigChecker(name string, configfiles ...string) (*configChecker, error
...
+// if reload equals to false, config checker won't update the checksum value in cahce, and function Check would return error forever if configfiles been modified.
```
qiujian16: if configfiles is modified.

pkg/utils/config_checker.go
xuezhaojun: Done.

pkg/utils/config_checker.go
xuezhaojun: Done

qiujian16 (commented):
/approve
/lgtm

PR #42, opened 2021-10-15
add certrotation from registration-operator to addon-framework

xuezhaojun:

### DIFF
--- a/pkg/certrotation/cabundle.go
+++ b/pkg/certrotation/cabundle.go
@@ -1,0 +1,109 @@
+package certrotation
+
+import (
+	"context"
+	"crypto/x509"
+	"fmt"
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
+
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+)
+
+// CABundleRotation maintains a CA bundle config map, but adding new CA certs and removing expired old ones.
+type CABundleRotation struct {
+	Namespace     string
+	Name          string
+	Lister        corev1listers.ConfigMapLister
+	Client        corev1client.ConfigMapsGetter
+	EventRecorder events.Recorder
+}
+
+func (c CABundleRotation) EnsureConfigMapCABundle(signingCertKeyPair *crypto.CA) ([]*x509.Certificate, error) {
+	// by this point we have current signing cert/key pair.  We now need to make sure that the ca-bundle configmap has this cert and
+	// doesn't have any expired certs
+	originalCABundleConfigMap, err := c.Lister.ConfigMaps(c.Namespace).Get(c.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+	caBundleConfigMap := originalCABundleConfigMap.DeepCopy()
+	if apierrors.IsNotFound(err) {
+		// create an empty one
+		caBundleConfigMap = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: c.Namespace, Name: c.Name}}
+	}
+	if _, err = manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0]); err != nil {
+		return nil, err
+	}
+	if originalCABundleConfigMap == nil || originalCABundleConfigMap.Data == nil || !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
+		c.EventRecorder.Eventf("CABundleUpdateRequired", "%q in %q requires update", c.Name, c.Namespace)
+		actualCABundleConfigMap, _, err := resourceapply.ApplyConfigMap(context.TODO(), c.Client, c.EventRecorder, caBundleConfigMap)
+		if err != nil {
+			return nil, err
+		}
+		caBundleConfigMap = actualCABundleConfigMap
+	}
+
+	caBundle := caBundleConfigMap.Data["ca-bundle.crt"]
+	if len(caBundle) == 0 {
+		return nil, fmt.Errorf("configmap/%s -n%s missing ca-bundle.crt", caBundleConfigMap.Name, caBundleConfigMap.Namespace)
+	}
+	certificates, err := cert.ParseCertsPEM([]byte(caBundle))
+	if err != nil {
+		return nil, err
+	}
+
+	return certificates, nil
+}
+
+// manageCABundleConfigMap adds the new certificate to the list of cabundles, eliminates duplicates, and prunes the list of expired
+// certs to trust as signers
+func manageCABundleConfigMap(caBundleConfigMap *corev1.ConfigMap, currentSigner *x509.Certificate) ([]*x509.Certificate, error) {
+	if caBundleConfigMap.Data == nil {
+		caBundleConfigMap.Data = map[string]string{}
+	}
+
+	certificates := []*x509.Certificate{}
+	caBundle := caBundleConfigMap.Data["ca-bundle.crt"]
+	if len(caBundle) > 0 {
+		var err error
+		certificates, err = cert.ParseCertsPEM([]byte(caBundle))
+		if err != nil {
+			return nil, err
+		}
+	}
+	certificates = append([]*x509.Certificate{currentSigner}, certificates...)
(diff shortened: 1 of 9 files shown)


pkg/certrotation/target.go
```diff
@@ -0,0 +1,141 @@
...
+	Namespace     string
+	Name          string
+	Validity      time.Duration
+	HostNames     []string
+	Lister        corev1listers.SecretLister
+	Client        corev1client.SecretsGetter
+	EventRecorder events.Recorder
+}
+
+func (c TargetRotation) EnsureTargetCertKeyPair(signingCertKeyPair *crypto.CA, caBundleCerts []*x509.Certificate) error {
```
yue9944882, line 36: where is this method referenced? i can clearly see the purpose and the implementation of this method but didnt find anywhere using it.

pkg/certrotation/cabundle.go
```diff
@@ -0,0 +1,109 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
+
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+)
+
+// CABundleRotation maintains a CA bundle config map, but adding new CA certs and removing expired old ones.
```
yue9944882, line 22: a bit curious why is it stored in the configmap instead of secret?

pkg/certrotation/target.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	cert := certificates[0]
+	if time.Now().After(cert.NotAfter) {
+		return "already expired"
+	}
+
+	maxWait := cert.NotAfter.Sub(cert.NotBefore) / 5
+	latestTime := cert.NotAfter.Add(-maxWait)
+	if time.Now().After(latestTime) {
```
yue9944882, line 93: https://github.com/open-cluster-management-io/registration/blob/ef7c23ededb2f886fc5e799d2125637709ba1688/pkg/clientcert/cert_controller.go#L199

it should be jittered so that certs can be rotated progressively, otherwise all the old certs will be wiped out at the same moment (which is practically dangerous)?

pkg/certrotation/target.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Sprintf("expired in %6.3f seconds", cert.NotAfter.Sub(time.Now()).Seconds())
+	}
+
+	// check the signer common name against all the common names in our ca bundle so we don't refresh early
```
yue9944882, line 97: it can be a dumb question but i'm confused 🤔 why does the CN of the ca bundles have to be the same w/ the [server|client] certificate?

yue9944882 (commented):
overall i think this will be a really useful feature for those addon developers who're looking for a lightweight certificate mgmt support in OCM. otherwise we will have to involve external components such as `cert-manager` which brings additional cost&complexity in hub cluster. e.g. the cluster-proxy addon can sign server certificates for proxy-servers and in the future i hope it can sign server certs for ingress-controllers.

pkg/certrotation/target.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	return nil
+}
+
+func (c TargetRotation) NewCertificate(signer *crypto.CA, validity time.Duration) (*crypto.TLSCertificateConfig, error) {
+	if len(c.HostNames) == 0 {
+		return nil, fmt.Errorf("no hostnames set")
+	}
+	return signer.MakeServerCertForDuration(sets.NewString(c.HostNames...), validity)
```
yue9944882, line 140: nit: i clicked into this function and it seems that it will automatically discriminate dnsname or ip address for us,  is it better if we can clarify it's "hostname or ip" in the variable name and the error messages?

pkg/certrotation/target.go
```diff
@@ -0,0 +1,141 @@
...
+func (c TargetRotation) setTargetCertKeyPairSecret(targetCertKeyPairSecret *corev1.Secret, validity time.Duration, signer *crypto.CA) error {
+	if targetCertKeyPairSecret.Data == nil {
+		targetCertKeyPairSecret.Data = map[string][]byte{}
+	}
+
+	// make sure that we don't specify something past our signer
+	targetValidity := validity
+	// TODO: When creating a certificate, crypto.MakeServerCertForDuration accetps validity as input parameter,
+	// It calls time.Now() as the current time to calculate NotBefore/NotAfter of new certificate, which might
+	// be little later than the returned value of time.Now() call in the line below to get remainingSignerValidity.
```
yue9944882, line 117: curiously any negative things will happen if the signed certificates' `NotBefore/NotAfter` lags ~1 sec behind?

pkg/certrotation/target.go
```diff
@@ -0,0 +1,141 @@
...
+	"crypto/x509"
+	"fmt"
+	"time"
+
+	"github.com/openshift/library-go/pkg/certs"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
```
yue9944882, line 14: nit: remove the blank line or sort the imports by their package names

xuezhaojun:
@yue9944882 Hello Kim, Thanks for the view. This new add pkg is actually a copy from registration-operator [certrotation](https://github.com/open-cluster-management-io/registration-operator/tree/main/pkg/certrotation). The motivation is we are working on some addons needs this feature and currently they leverage certroatation from registration-operator directly and it's not a proper way to do this.

So the first idea I come up with is to move this pkg to addon-framework to eliminate the dependency of registration-operator.  But I haven't gone to the details yet honestly so I may not enable to answer all question immediately.

I can use a couple of days to comprehend this pkg and then answers the above questions more accurately. Meanwhile, welcome any new ideas, requirements or designs for this feature.

pkg/certrotation/cabundle.go
qiujian16: CA is public so it is not necessary to be stored in secret

pkg/certrotation/target.go
qiujian16: it is not necessary I think

pkg/certrotation/cabundle.go
yue9944882: yeah, we store ca in configmap for "kube-system/extension-apiserver-authentication" too

pkg/certrotation/target.go
xuezhaojun: imports formated

pkg/certrotation/target.go
xuezhaojun: This method is involved by the client code.

pkg/certrotation/target.go
xuezhaojun: Here is a example of usage:
```
for _, targetRotation := range c.targetRotations {
		if err := targetRotation.EnsureTargetCertKeyPair(signingCertKeyPair, cabundleCerts); err != nil {
			errs = append(errs, err)
		}
	}
```
We can see that `EnsureTargetCertKeyPair` is invoked one by one, so as `needNewTargetCertKeyPair` and `setTargetCertKeyPairSecret`. 
So actually it won't update all old certs at the same time.

pkg/certrotation/target.go
xuezhaojun: For this part, we may end up with a inconsistent moment that the singer is rotated but the signed certificates are not.

pkg/certrotation/target.go
xuezhaojun: I think you means this funciton in library-go:
```
func IPAddressesDNSNames(hosts []string) ([]net.IP, []string) {
	ips := []net.IP{}
	dns := []string{}
	for _, host := range hosts {
		if ip := net.ParseIP(host); ip != nil {
			ips = append(ips, ip)
		} else {
			dns = append(dns, host)
		}
	}

	// Include IP addresses as DNS subjectAltNames in the cert as well, for the sake of Python, Windows (< 10), and unnamed other libraries
	// Ensure these technically invalid DNS subjectAltNames occur after the valid ones, to avoid triggering cert errors in Firefox
	// See https://bugzilla.mozilla.org/show_bug.cgi?id=1148766
	for _, ip := range ips {
		dns = append(dns, ip.String())
	}

	return ips, dns
}
```

For now this cert pkg depends on `openshift/libaray-go/crypto`, but maybe later we could make a bigger refactor to replace this dependency with a standard `crypto` lib and solve this.

pkg/certrotation/target.go
xuezhaojun: It checks the "sign" relationship between the signer and signed certificates. It could happen that a valid target cert is not signed by the current signer.

The signer and caBundle could be updated already(and the new caBundle is sent to customer/client). Then a targetCert needs to use this check to be aware of the signer's update and rotate itself.

xuezhaojun:
Hello, @qiujian16 @yue9944882 I have replied to the comments, I think we can collect more use cases from addons in the future and do a refactor then.

yue9944882:
@qiujian16 am +1 to get this merged in the current phase, hopefully i will be getting back w/ more feedbacks after rebasing our addons onto it.

PR #51, opened 2021-12-09
Feat: Factor out a health prober option for addons

yue9944882:
> Signed-off-by: yue9944882 <291271447@qq.com>

/cc @qiujian16


pkg/addonmanager/controllers/addoninstall/controller.go
```diff
@@ -107,19 +107,49 @@ func (c *addonInstallController) applyAddon(ctx context.Context, addonName, clus
...
+		if err != nil {
+			return err
+		}
 		return nil
 	}
 
-	addon = addon.DeepCopy()
-	addon.Spec.InstallNamespace = installNamespace
-	_, err = c.addonClient.AddonV1alpha1().ManagedClusterAddOns(clusterName).Update(ctx, addon, metav1.UpdateOptions{})
+	// reconcile health check mode
```
qiujian16: I think the status tracking should be independent with install addon. If an addon is not auto installed, it should still be able to define a health checker to update status. It might be put in a separated controller.

yue9944882: separated an addon health check controller

qiujian16 (commented):
/approve
/lgtm

PR #53, opened 2021-12-14
Feat: Apply an additional x509 parent check when looking for issuer

yue9944882:
> Signed-off-by: yue9944882 <291271447@qq.com>

/cc @qiujian16


qiujian16 (commented):
/approve
/lgtm

PR #57, opened 2021-12-17
implement an addon template using embed.FS

zhiweiyin318:
1. create an agentAddon with options, and implement its manifest function using embed.FS.
2. can override and customize the template config by the annotation of manangedClusterAddon CR.
issue : https://github.com/open-cluster-management-io/community/issues/117


pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,138 @@
...
+	case AgentAddonCustomMode:
+		factory.agentAddon = nil
+	default:
+		factory.agentAddon = newDefaultAgentAddon(factory)
+	}
+
+	return factory
+}
+
+func (a *addonFactory) Start(ctx context.Context) {
```
qiujian16: no need for this, you could have a build() func that return an agent interface.

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,138 @@
...
+	if a.agentAddon == nil {
+		klog.Exit(fmt.Errorf("invalid agentAddon defined"))
+	}
+
+	a.addonManager.AddAgent(a.agentAddon)
+	a.addonManager.Start(ctx)
+}
+
+func WithInstallStrategy(installAll bool, installNamespace string) AddonOptions {
```
qiujian16: Why not `WithInstallStrategy(installAll bool, installNamespace string) *addonFactory`

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,138 @@
...
+				Type:             agent.InstallAll,
+				InstallNamespace: installNamespace,
+			}
+		}
+
+		return factory
+	}
+}
+
+func WithPermissionConfig(permissionManifests ...runtime.Object) AddonOptions {
```
qiujian16: I think you could have an `WithAgentAddonOption(o *AgentAddonOptions)*addonFactory` If a user needs additional part for registration

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,138 @@
...
+	return func(factory *addonFactory) *addonFactory {
+		factory.permissionManifests = permissionManifests
+		return factory
+	}
+}
+
+func WithStaticManifests(staticManifests ...runtime.Object) AddonOptions {
```
qiujian16: I will not be easy to do this..how about `func WithStaticManifests(staticManifests ...[]byte) *addonFactory` And in the build() func you should validate all the manifests

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,138 @@
...
+	return func(factory *addonFactory) *addonFactory {
+		factory.staticManifests = staticManifests
+		return factory
+	}
+}
+
+func WithTemplateManifests(templateManifests TemplateManifests) AddonOptions {
```
qiujian16: `func WithTemplateManifests(config interface{}, templateManifests ...[]byte) *addonFactory`?

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,138 @@
...
+	return func(factory *addonFactory) *addonFactory {
+		factory.templateManifests = templateManifests
+		return factory
+	}
+}
+
+func WithAgentAddon(agentAddon agent.AgentAddon) AddonOptions {
```
qiujian16: why need this?

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,75 @@
...
+	f.scheme = scheme
+	return f
+}
+
+func (f *AgentAddonFactory) WithTemplateConfig(config map[string]interface{}) *AgentAddonFactory {
+	f.templateConfig = config
+	return f
+}
+
+func (f *AgentAddonFactory) WithInstallStrategy(installAll bool, installNamespace string) *AgentAddonFactory {
```
qiujian16: I think you it is better to do it like

```
WithInstallStrategy(type agent.installType, options ...strategyOption)
```

It makes your func signature more extensible

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,75 @@
...
+	return f
+}
+
+func (f *AgentAddonFactory) WithAgentRegistrationOption(option *agent.RegistrationOption) *AgentAddonFactory {
+	f.agentAddonOptions.Registration = option
+	return f
+}
+
+func (f *AgentAddonFactory) Build() agent.AgentAddon {
+	if f.scheme == nil {
```
qiujian16: I would think you need to do at least validate the manifest here, and return error if anything is wrong

pkg/addonfactory/factory.go
zhiweiyin318: currently InstallStrategy only supports InstallAll type. changed to 
` WithInstallStrategy(installType agent.StrategyType, installNamespace string)`
I have no idea how to define `strategyOption`.
I think we can add another function like `WithInstallStrategyOption` in the future when we  extend more types of  InstallStrategy. 
any suggestion?

pkg/addonfactory/factory.go
zhiweiyin318: added validation of template render and object decoder.

pkg/addonfactory/defaultagentaddon.go
```diff
@@ -0,0 +1,128 @@
...
+	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+// Template config name in annotation
+const templateConfigName string = "templateConfig"
+
+// Template field name
+const (
+	ClusterName           string = "ClusterName"
```
qiujian16: should this be public?

pkg/addonfactory/defaultagentaddon.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	manifestConfig[ClusterName] = cluster.GetName()
+
+	installNamespace := addon.Spec.InstallNamespace
+	if len(installNamespace) == 0 {
+		installNamespace = AddonDefaultInstallNamespace
+	}
+	manifestConfig[AddonInstallNamespace] = installNamespace
+	manifestConfig[KubeConfigSecret] = fmt.Sprintf("%s-hub-kubeconfig", a.agentAddonOptions.AddonName)
```
qiujian16: this depends on whether addon enables registration i think

pkg/addonfactory/defaultagentaddon.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// get templateConfig from annotation of addon cr
+	annotations := addon.GetAnnotations()
```
qiujian16: why we need this? and this needs to be well documented.

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,91 @@
...
+// the reserved fields (ClusterName,AddonInstallNamespace and KubeConfigSecret) are not mandatory in the config.
+func (f *AgentAddonFactory) WithTemplateConfig(config interface{}) *AgentAddonFactory {
+	f.templateConfig = config
+	return f
+}
+
+// WithInstallStrategy defines the installation strategy of the manifests prescribed by Manifests(..).
+// installType is the type of the installation strategy. currently only supported values: "agent.InstallAll".
+// installNamespace is target deploying namespace in the managed cluster upon automatic addon installation.
+func (f *AgentAddonFactory) WithInstallStrategy(installType agent.StrategyType, installNamespace string) *AgentAddonFactory {
```
qiujian16: maybe set InstallStrategy, otherwise it will be hard to evolve

pkg/addonfactory/factory.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"open-cluster-management.io/addon-framework/pkg/agent"
+)
+
+const AddonDefaultInstallNamespace = "open-cluster-management-agent-addon"
+
+type AgentAddonFactory struct {
```
qiujian16: comment on this

qiujian16 (commented):
/assign @yue9944882 

would you also take a look?

pkg/addonfactory/factory.go
zhiweiyin318: done.

pkg/addonfactory/defaultagentaddon.go
zhiweiyin318: doced.

pkg/addonfactory/defaultagentaddon.go
zhiweiyin318: the KubeConfigSecret is defined as a reserved filed, it will not be used if addon is not registration.

pkg/addonfactory/defaultagentaddon.go
zhiweiyin318: changed to private.

pkg/addonfactory/factory.go
qiujian16: how about `WithInstallStrategyOption(DefaultInstallAllOption)`, you could provide a default install strategy leaving the namespace empty.

pkg/addonfactory/defaultagentaddon.go
qiujian16: I think you have to check different scenario and generate different var here. You could have multiple signers defined in the addon and some signers may not be client cert signer

yue9944882 (commented):
it's great to use go embed to build the manifests into the binary!  in addtion to the pr i think we can provide similar support for managing the rbac manifests applied by the permission-setup callback. e.g. currently in the clusterproxy addon we didnt actively updates the applied rbac policies, we're staying at "create-if-not-exists" for simplicity. also i  have a few minor questions about the sustainability of maintaining these templates:

- should we inject owner-references to the corresponding managedclusteraddon so that they can be automatically removed upon un-installation?
- plz clarify the order of the applied items, considering there can be cross-references/depedencies between them.
- plz clarify the failure model upon applying illegal template. e.g. what if a malformed yaml (e.g. w/ illegal indent, multiple resources, empty file etc.) is embed, are we expected to see the failure via events or the status of the managedclusteraddon (merely throwing an error personally doesnt SGTM tho)

qiujian16:
@yue9944882 we are also thinking alternatives like allow user to put a helm chart to define agent for addon. Where we can run helm dryrun to get the manifests and put in the manifestwork. WDYT?

yue9944882:
@qiujian16 that will be helpful for those who trying to migrate the manifests from helm charts to an addon, so addon developers can build an addon via adding the chart as a git submodule in their project. but we can start w the low hanging fruit by copying/duplicating the manifests to the addon project. 

i presume the complexity of integrating helm chart lies in the advanced templating extensions provided by charts e.g. flow-control, pre|post-jobs, etc. i am a bit worried if these charts w advanced annotation can always be correctly rendered.

pkg/addonfactory/defaultagentaddon.go
qiujian16: let's remove this part, but let user to define something similar like what we do with values in helmagent

pkg/addonfactory/addonfactory.go
```diff
@@ -0,0 +1,35 @@
...
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+const AddonDefaultInstallNamespace = "open-cluster-management-agent-addon"
+
+// AnnotationValuesName is the annotation Name of customized values
+const AnnotationValuesName string = "addon.open-cluster-management.io/values"
+
+// BuiltinValues includes the built-in values
+// for helm chart, the values should begin with a lowercase letter, so we need convert it to Values by JsonStructToValues.
```
qiujian16: I'd rather we do not use the same struct for both helm and template.

pkg/addonfactory/helmaddonfactory/helm_addonfactory.go
```diff
@@ -42,30 +41,30 @@ func NewAgentAddonFactoryWithHelmChartFS(addonName string, fs embed.FS, chartPre
...
-func (f *HelmAgentAddonFactory) WithScheme(scheme *runtime.Scheme) *HelmAgentAddonFactory {
+func (f *HelmAgentAddonFactory) WithScheme(scheme *runtime.Scheme) addonfactory.AgentAddonFactory {
 	f.scheme = scheme
 	return f
 }
 
 // WithGetValuesFuncs adds a list of the getValues func.
 // the values got from the big index Func will override the one from small index Func.
-func (f *HelmAgentAddonFactory) WithGetValuesFuncs(getValuesFuncs []GetValuesFunc) *HelmAgentAddonFactory {
+func (f *HelmAgentAddonFactory) WithGetValuesFuncs(getValuesFuncs []addonfactory.GetValuesFunc) addonfactory.AgentAddonFactory {
```
qiujian16: WithGetValuesFuncs(getValuesFuncs ...addonfactory.GetValuesFunc)

zhiweiyin318: done

pkg/addonfactory/addonfactory.go
zhiweiyin318: done. defined different struct for helm and template

qiujian16 (commented):
/approve

PR #59, opened 2021-12-21
Add csr helpers func

qiujian16:

### DIFF
--- a/pkg/utils/csr_helpers.go
+++ b/pkg/utils/csr_helpers.go
@@ -1,0 +1,180 @@
+package utils
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"math/big"
+	"strings"
+	"time"
+
+	certificatesv1 "k8s.io/api/certificates/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+	"open-cluster-management.io/addon-framework/pkg/agent"
+	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+var serialNumberLimit = new(big.Int).Lsh(big.NewInt(1), 128)
+
+// DefaultSignerWithExpiry generates a signer func for addon agent to sign the csr using caKey and caData with expiry date.
+func DefaultSignerWithExpiry(caKey, caData []byte, duration time.Duration) agent.CSRSignerFunc {
+	return func(csr *certificatesv1.CertificateSigningRequest) []byte {
+		blockTlsCrt, _ := pem.Decode(caData) // note: the second return value is not error for pem.Decode; it's ok to omit it.
+		certs, err := x509.ParseCertificates(blockTlsCrt.Bytes)
+		if err != nil {
+			klog.Errorf("Failed to parse cert: %v", err)
+			return nil
+		}
+
+		blockTlsKey, _ := pem.Decode(caKey)
+
+		// For now only PKCS#1 is supported which assures the private key algorithm is RSA.
+		// TODO: Compatiblity w/ PKCS#8 key e.g. EC algorithm
+		key, err := x509.ParsePKCS1PrivateKey(blockTlsKey.Bytes)
+		if err != nil {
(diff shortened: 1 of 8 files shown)


qiujian16:
/hold
still need to add unit test

xuezhaojun:
/approve 
Looks good to me.

pkg/utils/csr_helpers.go
```diff
@@ -0,0 +1,144 @@
...
+	"k8s.io/klog/v2"
+	"open-cluster-management.io/addon-framework/pkg/agent"
+	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+// DefaultSigner generates a signer func for addon agent to sign the csr using caKey and caData.
+func DefaultSigner(caKey, caData []byte) agent.CSRSignerFunc {
+	return func(csr *certificatesv1.CertificateSigningRequest) []byte {
+		blockTlsCrt, _ := pem.Decode(caData) // note: the second return value is not error for pem.Decode; it's ok to omit it.
```
yue9944882, line 27: > // note: the second return value is not error for pem.Decode; it's ok to omit it.
Write Preview

can you elaborate why the error is neglectable? what if the passing caData is not a valid base64 data?

pkg/utils/csr_helpers.go
```diff
@@ -0,0 +1,144 @@
...
+		certs, err := x509.ParseCertificates(blockTlsCrt.Bytes)
+		if err != nil {
+			klog.Errorf("Failed to parse cert: %v", err)
+			return nil
+		}
+
+		blockTlsKey, _ := pem.Decode(caKey)
+		key, err := x509.ParsePKCS1PrivateKey(blockTlsKey.Bytes)
```
yue9944882, line 38: lets clarify in the comment that:

```
// For now only PKCS#1 is supported which assures the private key algorithm is RSA.
// TODO: Compatiblity w/ PKCS#8 key e.g. EC algorithm
```

pkg/utils/csr_helpers.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+func signCSR(csr *certificatesv1.CertificateSigningRequest, caCert *x509.Certificate, caKey *rsa.PrivateKey) ([]byte, error) {
+	var usages []string
+	for _, usage := range csr.Spec.Usages {
+		usages = append(usages, string(usage))
+	}
+
+	certExpiryDuration := 365 * 24 * time.Hour
```
yue9944882: how about adding another public method `DefaultSignerWithExpiry` that allows custom expiry duration?

pkg/utils/csr_helpers.go
xuezhaojun:
``` golang
// Decode will find the next PEM formatted block (certificate, private key
// etc) in the input. It returns that block and the remainder of the input. If
// no PEM data is found, p is nil and the whole of the input is returned in
// rest.
func Decode(data []byte) (p *Block, rest []byte) {
...
```
The second return value is `[]byte` type which we don't need in next steps. And If caData is not a valid base64 data, then the `blockTlsCrt` would be nil.

pkg/utils/csr_helpers.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	policy := &config.Signing{
+		Default: &config.SigningProfile{
+			Usage:        usages,
+			Expiry:       certExpiryDuration,
+			ExpiryString: certExpiryDuration.String(),
+		},
+	}
+
+	cfs, err := local.NewSigner(caKey, caCert, signer.DefaultSigAlgo(caKey), policy)
```
yue9944882: personally, am not fan of involving external (cfssl) dependency to sign the certificate b/c they're already available in the go sdk. consider following the sample [here](https://github.com/open-cluster-management-io/cluster-proxy/blob/297e092ffcb13fbc7192b1d7b6c1dd2c8f4f15e7/pkg/addon/agent/agent.go#L128-L173)

pkg/utils/csr_helpers.go
yue9944882: oh i thought its an error

pkg/utils/csr_helpers.go
qiujian16: yes, just change to DefaultSignerWithExpiry.

yue9944882 (commented):
/lgtm

PR #64, opened 2022-01-17
Upgrade go and library

qiujian16:


zhiweiyin318:
the same addon-agents create the same lease in default ns https://github.com/open-cluster-management-io/addon-framework/blob/10147aa1fdcfb56a9486172cbaeee14533f65908/examples/helloworld/helloworldagent.go#L87
so the helm-addon is not available

PR #71, opened 2022-01-27
 add pre-delete hook manifest

zhiweiyin318:
task: https://github.com/open-cluster-management-io/community/issues/119

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/controller.go
+++ b/pkg/addonmanager/controllers/agentdeploy/controller.go
@@ -10,7 +10,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -93,6 +92,9 @@ func NewAddonDeployController(
 				if _, ok := c.agentAddons[addonName]; !ok {
 					return false
 				}
+				if accessor.GetName() != deployWorkName(addonName) {
+					return false
+				}
 				return true
 			},
 			workInformers.Informer(),
@@ -118,7 +120,7 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 	// Get ManagedCluster
 	managedCluster, err := c.managedClusterLister.Get(clusterName)
 	if errors.IsNotFound(err) {
-		c.cache.removeCache(addonName, clusterName)
+		c.cache.removeCache(deployWorkName(addonName), clusterName)
 		return nil
 	}
 	if err != nil {
@@ -132,7 +134,7 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 
 	managedClusterAddon, err := c.managedClusterAddonLister.ManagedClusterAddOns(clusterName).Get(addonName)
 	if errors.IsNotFound(err) {
-		c.cache.removeCache(addonName, clusterName)
+		c.cache.removeCache(deployWorkName(addonName), clusterName)
 		return nil
 	}
 	if err != nil {
@@ -140,7 +142,7 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 	}
 
 	if !managedClusterAddon.DeletionTimestamp.IsZero() {
-		c.cache.removeCache(addonName, clusterName)
+		c.cache.removeCache(deployWorkName(addonName), clusterName)
 		return nil
 	}
 
@@ -155,14 +157,14 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 		return nil
 	}
 
-	work, err := buildManifestWorkFromObject(clusterName, addonName, objects)
+	work, _, err := buildManifestWorkFromObject(clusterName, addonName, objects)
 	if err != nil {
 		return err
 	}
 	work.OwnerReferences = []metav1.OwnerReference{*owner}
 
 	// apply work
-	work, err = c.applyWork(ctx, work)
+	work, err = applyWork(c.workClient, c.workLister, c.cache, c.eventRecorder, ctx, work)
 	if err != nil {
 		meta.SetStatusCondition(&managedClusterAddonCopy.Status.Conditions, metav1.Condition{
 			Type:    "ManifestApplied",
@@ -199,82 +201,4 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 	_, err = c.addonClient.AddonV1alpha1().ManagedClusterAddOns(managedClusterAddonCopy.Namespace).UpdateStatus(
 		ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
 	return err
-}
-
-func (c *addonDeployController) applyWork(ctx context.Context, required *workapiv1.ManifestWork) (*workapiv1.ManifestWork, error) {
-	existingWork, err := c.workLister.ManifestWorks(required.Namespace).Get(required.Name)
-	existingWork = existingWork.DeepCopy()
-	if errors.IsNotFound(err) {
-		existingWork, err = c.workClient.WorkV1().ManifestWorks(required.Namespace).Create(ctx, required, metav1.CreateOptions{})
-		if err == nil {
-			c.eventRecorder.Eventf("ManifestWorkCreated", "Created %s/%s because it was missing", required.Namespace, required.Name)
-			c.cache.updateCache(required, existingWork)
-			return existingWork, nil
-		}
-		c.eventRecorder.Warningf("ManifestWorkCreateFailed", "Failed to create ManifestWork %s/%s: %v", required.Namespace, required.Name, err)
-		return nil, err
-	}
-	if err != nil {
-		return nil, err
-	}
-
-	if c.cache.safeToSkipApply(required, existingWork) {
-		return existingWork, nil
-	}
-
-	if ManifestsEqual(required.Spec.Workload.Manifests, existingWork.Spec.Workload.Manifests) {
-		return existingWork, nil
-	}
-
-	existingWork.Spec.Workload = required.Spec.Workload
-	existingWork, err = c.workClient.WorkV1().ManifestWorks(existingWork.Namespace).Update(ctx, existingWork, metav1.UpdateOptions{})
-	if err == nil {
-		c.cache.updateCache(required, existingWork)
-		c.eventRecorder.Eventf("ManifestWorkUpdate", "Updated %s/%s because it was changing", required.Namespace, required.Name)
-		return existingWork, nil
-	}
-	c.eventRecorder.Warningf("ManifestWorkUpdateFailed", "Failed to update ManifestWork %s/%s: %v", required.Namespace, required.Name, err)
-	return nil, err
-}
-
-func buildManifestWorkFromObject(cluster, addonName string, objects []runtime.Object) (*workapiv1.ManifestWork, error) {
-	work := &workapiv1.ManifestWork{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("addon-%s-deploy", addonName),
-			Namespace: cluster,
-			Labels: map[string]string{
-				constants.AddonLabel: addonName,
-			},
-		},
-		Spec: workapiv1.ManifestWorkSpec{
-			Workload: workapiv1.ManifestsTemplate{
-				Manifests: []workapiv1.Manifest{},
-			},
-		},
-	}
-
-	for _, object := range objects {
-		rawObject, err := runtime.Encode(unstructured.UnstructuredJSONScheme, object)
-		if err != nil {
-			return nil, err
-		}
-		work.Spec.Workload.Manifests = append(work.Spec.Workload.Manifests, workapiv1.Manifest{
-			RawExtension: runtime.RawExtension{Raw: rawObject},
-		})
-	}
-
-	return work, nil
-}
-
-func ManifestsEqual(new, old []workapiv1.Manifest) bool {
-	if len(new) != len(old) {
-		return false
-	}
-
-	for i := range new {
-		if !equality.Semantic.DeepEqual(new[i].Raw, old[i].Raw) {
-			return false
-		}
-	}
-	return true
 }
--- a/pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
+++ b/pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
@@ -1,0 +1,277 @@
+package agentdeploy
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	"open-cluster-management.io/addon-framework/pkg/addonmanager/constants"
+	"open-cluster-management.io/addon-framework/pkg/agent"
+	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	addonv1alpha1client "open-cluster-management.io/api/client/addon/clientset/versioned"
+	addoninformerv1alpha1 "open-cluster-management.io/api/client/addon/informers/externalversions/addon/v1alpha1"
+	addonlisterv1alpha1 "open-cluster-management.io/api/client/addon/listers/addon/v1alpha1"
+	clusterinformers "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterlister "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	workv1client "open-cluster-management.io/api/client/work/clientset/versioned"
+	workinformers "open-cluster-management.io/api/client/work/informers/externalversions/work/v1"
+	worklister "open-cluster-management.io/api/client/work/listers/work/v1"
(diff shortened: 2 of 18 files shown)


pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -131,52 +131,86 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
+		if !hasFinalizer(managedClusterAddonCopy.GetFinalizers(), constants.PreDeleteHookFinalizer) {
+			managedClusterAddonCopy.Finalizers = append(managedClusterAddonCopy.Finalizers, constants.PreDeleteHookFinalizer)
+			_, err = c.addonClient.AddonV1alpha1().ManagedClusterAddOns(managedClusterAddonCopy.Namespace).Update(
+				ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
+			return err
+		}
+	}
+
+	if !managedClusterAddon.DeletionTimestamp.IsZero() {
+		c.cache.removeCache(addonName, clusterName)
```
qiujian16: should hookWork also be put in the cache?

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -131,52 +131,86 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
+				return err
+			}
+
+			if hookWorkIsFinished(hookWork) {
+				finalizer := removeFinalizer(managedClusterAddonCopy.Finalizers, constants.PreDeleteHookFinalizer)
+				managedClusterAddonCopy.SetFinalizers(finalizer)
+				_, err = c.addonClient.AddonV1alpha1().ManagedClusterAddOns(managedClusterAddonCopy.Namespace).Update(
+					ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
+				return err
+			}
```
qiujian16: I think we also need another condition here to indicate that addon is pending in deletion since some hook is not done yet.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -131,52 +131,86 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
+			return fmt.Errorf("reconcile since the hook manifetsWorks is not finished")
```
qiujian16: I do not think we should return error here. We should monitor hookWork, and reconcile when hookwork is updated.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -237,44 +271,202 @@ func (c *addonDeployController) applyWork(ctx context.Context, required *workapi
...
-	for _, object := range objects {
-		rawObject, err := runtime.Encode(unstructured.UnstructuredJSONScheme, object)
-		if err != nil {
-			return nil, err
+func newManifestConfigOption(obj *unstructured.Unstructured) workapiv1.ManifestConfigOption {
+	var resource string
+	var jsonPath workapiv1.JsonPath
+	gvk := obj.GroupVersionKind()
+	switch gvk.Kind {
+	case "Job":
```
qiujian16: I think we should move this as part of WellKnownStatus in work agent

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -237,44 +271,202 @@ func (c *addonDeployController) applyWork(ctx context.Context, required *workapi
...
 			},
 		},
 	}
+}
 
-	for _, object := range objects {
-		rawObject, err := runtime.Encode(unstructured.UnstructuredJSONScheme, object)
-		if err != nil {
-			return nil, err
+func newManifestConfigOption(obj *unstructured.Unstructured) workapiv1.ManifestConfigOption {
```
qiujian16: add a comment on what is supported here.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -237,44 +271,202 @@ func (c *addonDeployController) applyWork(ctx context.Context, required *workapi
...
+				JsonPaths: []workapiv1.JsonPath{jsonPath},
+			},
+		},
+	}
+}
+func newHookManifestWork(
+	workName, addonName, clusterName string,
+	manifests []workapiv1.Manifest,
+	manifestConfigs []workapiv1.ManifestConfigOption) *workapiv1.ManifestWork {
+	hookWork := &workapiv1.ManifestWork{
```
qiujian16: I think you can call newDeployManifestWork here.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -237,44 +271,202 @@ func (c *addonDeployController) applyWork(ctx context.Context, required *workapi
...
+		},
 	}
 
-	return work, nil
+	return hookWork
 }
 
-func ManifestsEqual(new, old []workapiv1.Manifest) bool {
-	if len(new) != len(old) {
+func isPreDeleteHookObject(object *unstructured.Unstructured) bool {
```
qiujian16: this is doing a similar thing as newManifestConfigOption. You can have 

isPreDeleteHookObject(object *unstructured.Unstructured) (bool, *manifestConfigs)

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -237,44 +271,202 @@ func (c *addonDeployController) applyWork(ctx context.Context, required *workapi
...
+			return false
+		}
+		if *values[0].Value.String != "True" {
 			return false
 		}
 	}
 	return true
 }
+
+func (c *addonDeployController) runHookWork(
```
qiujian16: why not using applyWork here? If a hook is failed and we cannot update the manifest, How do we get out?

qiujian16 (commented):
We need ut and integration test for this. Especially the error case.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: handled following the deploywork

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: added  `HookManifestCompleted` condition.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: done.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: created another controller to do the pre delete jobs. and use the applyWork to apply hook manifests.

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,306 @@
...
+	addonlisterv1alpha1 "open-cluster-management.io/api/client/addon/listers/addon/v1alpha1"
+	clusterinformers "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterlister "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	workv1client "open-cluster-management.io/api/client/work/clientset/versioned"
+	workinformers "open-cluster-management.io/api/client/work/informers/externalversions/work/v1"
+	worklister "open-cluster-management.io/api/client/work/listers/work/v1"
+	workapiv1 "open-cluster-management.io/api/work/v1"
+)
+
+// managedClusterController reconciles instances of ManagedCluster on the hub.
```
qiujian16: update the comment

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,306 @@
...
+				ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
+			return err
+		}
+	}
+
+	if managedClusterAddon.DeletionTimestamp.IsZero() {
+		return nil
+	}
+	// apply hookWork when addon is deleting
+	if hookWork != nil {
```
qiujian16: I think you can do 

```
if hookWork == nil {
  return nil 
}
```
here?

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,306 @@
...
+				Reason:  "ManifestWorkApplyFailed",
+				Message: fmt.Sprintf("failed to apply manifestwork %v: %v", hookWork.Name, err),
+			})
+			c.addonClient.AddonV1alpha1().ManagedClusterAddOns(managedClusterAddonCopy.Namespace).UpdateStatus(
+				ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
+			return err
+		}
+
+		if !hookWorkIsCompleted(hookWork) {
+			meta.SetStatusCondition(&managedClusterAddonCopy.Status.Conditions, metav1.Condition{
```
qiujian16: Set condition False always and if work is completed, set it to true. So you would not have two deepequal in the code

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,306 @@
...
+	if errors.IsNotFound(err) {
+		c.cache.removeCache(hookWorkName(addonName), clusterName)
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if !managedCluster.DeletionTimestamp.IsZero() {
+		// managed cluster is deleting, do nothing
```
qiujian16: why not clean cache here?

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
qiujian16: I think you should alwasy removeFinalizer when hookWork == nil

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,306 @@
...
+	if !meta.IsStatusConditionTrue(hookWork.Status.Conditions, workapiv1.WorkAvailable) {
+		return false
+	}
+	for _, manifest := range hookWork.Status.ResourceStatus.Manifests {
+		values := manifest.StatusFeedbacks.Values
+		if len(values) == 0 {
+			return false
+		}
+		switch manifest.ResourceMeta.Resource {
+		case "jobs":
```
qiujian16, line 248: how about a helper func FindManifestValue(resourceStatus, identifier, valueName) value

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"open-cluster-management.io/addon-framework/pkg/addonmanager/constants"
+	workapiv1 "open-cluster-management.io/api/work/v1"
+)
+
+func deployWorkName(addonName string) string {
+	return fmt.Sprintf("addon-%s-deploy", addonName)
+}
+
+func hookWorkName(addonName string) string {
```
qiujian16: preDeleteHookWorkName?

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: added

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: added

pkg/addonmanager/controllers/agentdeploy/utils.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+		finalizer := removeFinalizer(managedClusterAddonCopy.Finalizers, constants.PreDeleteHookFinalizer)
+		managedClusterAddonCopy.SetFinalizers(finalizer)
+		_, err = c.addonClient.AddonV1alpha1().ManagedClusterAddOns(managedClusterAddonCopy.Namespace).Update(
+			ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
+		return err
+	}
+
+	// has hook manifestWork
+	if !hasFinalizer(managedClusterAddonCopy.GetFinalizers(), constants.PreDeleteHookFinalizer) {
```
qiujian16: should this be in if managedClusterAddon.DeletionTimestamp.IsZero() ?

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	meta.SetStatusCondition(&managedClusterAddonCopy.Status.Conditions, metav1.Condition{
+		Type:    "HookManifestCompleted",
+		Status:  metav1.ConditionFalse,
+		Reason:  "HookManifestIsNotCompleted",
+		Message: fmt.Sprintf("hook manifestWork %v is not completed.", hookWork.Name),
+	})
+
+	completed := hookWorkIsCompleted(hookWork)
```
qiujian16: I think you should check completed and remove finalizer  and return at first. It will remove the addon and the following status update will not be needed.

Then you can check whether there is an apply error or completed with switch.

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,301 @@
...
+		_, err = c.addonClient.AddonV1alpha1().ManagedClusterAddOns(managedClusterAddonCopy.Namespace).Update(
+			ctx, managedClusterAddonCopy, metav1.UpdateOptions{})
+		return err
+	}
+
+	return nil
+}
+
+func (c *addonHookDeployController) applyWork(ctx context.Context, required *workapiv1.ManifestWork) (*workapiv1.ManifestWork, error) {
+	existingWork, err := c.workLister.ManifestWorks(required.Namespace).Get(required.Name)
```
qiujian16: I think we should have a common apply func for both controllers.

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
qiujian16: something like

```go
hookWork, applyErr = c.applyWork(ctx, hookWork)
completed := hookWorkIsCompleted(hookWork)

if completed && hasFinalizer {
  return removeFinalizer()
}

switch {
case applyErr != nil:
case !completed
}

//deepEqual and update status
```

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if c.cache.safeToSkipApply(required, existingWork) {
+		return existingWork, nil
+	}
+
+	if manifestsEqual(required.Spec.Workload.Manifests, existingWork.Spec.Workload.Manifests) {
```
qiujian16: this will not be enough, since you still have manifestConfigs set.

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: fixed. 
yes, it will fail to add a finalizer to a deleting resource.

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: done. this is better. thanks.

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: changed to compare the whole spec in the common apply func.

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
```diff
@@ -0,0 +1,277 @@
...
+
+	if !managedCluster.DeletionTimestamp.IsZero() {
+		// managed cluster is deleting, remove cache.
+		c.cache.removeCache(preDeleteHookWorkName(addonName), clusterName)
+		return nil
+	}
+
+	managedClusterAddon, err := c.managedClusterAddonLister.ManagedClusterAddOns(clusterName).Get(addonName)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
qiujian16: change to 

```go
if errors.IsNotFound(err) {
			c.cache.removeCache(preDeleteHookWorkName(addonName), clusterName)
			return nil
}
if err != nil {
		return err
}
```

qiujian16 (commented):
/approve

pkg/addonmanager/controllers/agentdeploy/hookcontroller.go
zhiweiyin318: done

PR #72, opened 2022-01-27
Skip templates if 'Kind' not present

JustinKuli:
A template file might have a conditional block around the resource,
allowing it to be enabled/disabled based on other variables. Previously,
the controller could not handle this, it would error with a message like
```
controller failed to sync "...", err: Object 'Kind' is missing in '...'
```

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

### DIFF
--- a/pkg/addonfactory/helm_agentaddon.go
+++ b/pkg/addonfactory/helm_agentaddon.go
@@ -89,6 +89,10 @@ func (a *HelmAgentAddon) Manifests(
 		klog.V(4).Infof("%v/n", data)
 		object, _, err := a.decoder.Decode([]byte(data), nil, nil)
 		if err != nil {
+			if runtime.IsMissingKind(err) {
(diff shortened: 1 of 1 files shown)


JustinKuli:
/cc @zhiweiyin318 

Example template: https://github.com/JustinKuli/governance-policy-addon-controller/blob/f315f218e1e2b901e02848db16ce8bb78f9f7339/pkg/addon/policy_framework/manifests/managedcluster-chart/templates/policy.open-cluster-management.io_policies_crd.yaml - we don't want to deploy the CRD on the hub cluster, it would (usually) already be present from the other hub-side component of the policy framework.

Skipping the template works for the GRC case, but maybe this would be worth setting as a configuration option? So that it would fail to deploy in other use-cases.

zhiweiyin318:
/lgtm
thanks @JustinKuli

qiujian16 (commented):
/approve

This will need at least ut as a followup @zhiweiyin318

PR #73, opened 2022-01-28
synced fix to templateAgentAddon and add ut for them

zhiweiyin318:

### DIFF
--- a/pkg/addonfactory/template_agentaddon.go
+++ b/pkg/addonfactory/template_agentaddon.go
@@ -6,6 +6,7 @@ import (
 	"github.com/openshift/library-go/pkg/assets"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/klog/v2"
 	"open-cluster-management.io/addon-framework/pkg/agent"
 	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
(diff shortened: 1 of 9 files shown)


pkg/addonfactory/template_agentaddon.go
```diff
@@ -46,10 +48,27 @@ func (a *TemplateAgentAddon) Manifests(
 	if err != nil {
 		return objects, err
 	}
-	for file, data := range a.templateData {
-		raw := assets.MustCreateAssetFromTemplate(file, data, configValues).Data
+
+	// sort the filenames of the templates so the manifests are ordered consistently
+	keys := make([]string, 0, len(a.templateData))
```
qiujian16: why we need templateData to be a map initially?

zhiweiyin318: I get the data of template file from FS and template dir when build the agent, save the `file name : data` into the map.
the file name is used in `MustCreateAssetFromTemplate`.
otherwise, need save file name list and FS into the agentAddon struct, and get the data of template in `agentAddon.Manifests`.

zhiweiyin318: changed to save templateFiles in the struct, and use the templatefiles order to Iterate the templateData.

qiujian16: how about having a

```go
type templateFile struct {
  name string
  content []byte
}
```

zhiweiyin318: this definition is better. done.

qiujian16 (commented):
/approve
/lgtm

PR #76, opened 2022-02-16
Feat: Addon rbac permission utilities

yue9944882:


pkg/utils/permission.go
```diff
@@ -0,0 +1,275 @@
...
+			}
+		}
+		return nil
+	})
+	return p
+}
+
+func (p *permissionBuilder) WithStaticClusterRoleBinding(binding *rbacv1.ClusterRoleBinding) RBACPermissionBuilder {
+	p.u.fns = append(p.u.fns, func(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn) error {
+		ensureAddonOwnerReference(&binding.ObjectMeta, addon)
```
qiujian16: We cannot add this since this is a clusterrole

pkg/utils/permission.go
```diff
@@ -0,0 +1,275 @@
...
+					Kind: rbacv1.GroupKind,
+					Name: userGroup,
+				},
+			},
+		})
+}
+
+func (p *permissionBuilder) WithStaticClusterRole(clusterRole *rbacv1.ClusterRole) RBACPermissionBuilder {
+	p.u.fns = append(p.u.fns, func(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn) error {
+		ensureAddonOwnerReference(&clusterRole.ObjectMeta, addon)
```
qiujian16: We cannot add this since this is a clusterrole

pkg/utils/permission.go
```diff
@@ -0,0 +1,275 @@
...
+		exists := true
+		latest, err := p.kubeClient.RbacV1().RoleBindings(cluster.Name).Get(context.TODO(), binding.Name, metav1.GetOptions{})
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				return err
+			}
+			exists = false
+		}
+		if !exists {
+			_, err := p.kubeClient.RbacV1().RoleBindings(cluster.Name).Create(context.TODO(), binding, metav1.CreateOptions{})
```
qiujian16: use ApplyRole in library-go? I think we need to handle the rule change on rbac

pkg/utils/permission.go
yue9944882: removed ownerref

pkg/utils/permission.go
yue9944882: removed ownerref

pkg/utils/permission.go
yue9944882: good point, replace w `ApplyXXX`

qiujian16 (commented):
/approve
/lgtm

Issue #74, opened 2022-01-28
Use manifestwork status to update addon status


qiujian16:
Some agent use lease to maintain the status of managedclusteraddon. We should have an approach in addon-framework that let addon-controller to update available status of managedclusteraddon. It could be based on:

1. feedback status from manifestwork
2. other customized way.

qiujian16:
/kind feature

cc @yue9944882

yue9944882:
> feedback status from manifestwork

@qiujian16 mmm the idea is intriguing. i was planning to add work-based addon healthiness previously, but it's merely mapping the overall success of the work object to the healthiness of the addon. a finer-grained mapping will be def more helpful, we can start by converging their coding interface in the framework.

tesshuflower:
This feature would be very helpful for me, as I'm developing an addon controller that will not install an agent on managed clusters.

qiujian16:
this is done

/close

PR #77, opened 2022-02-23
Add work probe mode

qiujian16:

https://github.com/open-cluster-management-io/addon-framework/issues/74

### DIFF
--- a/pkg/agent/inteface.go
+++ b/pkg/agent/inteface.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	workapiv1 "open-cluster-management.io/api/work/v1"
 )
 
 // AgentAddon is a mandatory interface for implementing a custom addon.
@@ -117,6 +118,25 @@ type InstallStrategy struct {
 
 type HealthProber struct {
 	Type HealthProberType
+
+	WorkProber WorkHealthProber
+}
+
+type WorkHealthProber interface {
+	// ProbeFields tells addon framework what field to probe
+	ProbeFields() []ProbeField
+
+	// HealthCheck check status of the addon based on probe result.
+	HealthCheck(workapiv1.ResourceIdentifier, workapiv1.StatusFeedbackResult) error
+}
+
+// ProbeField defines the field of a resource to be probed
+type ProbeField struct {
+	// ResourceIdentifier sets what resource shoule be probed
+	ResourceIdentifier workapiv1.ResourceIdentifier
+
+	// ProbeRules sets the rules to probe the field
+	ProbeRules []workapiv1.FeedbackRule
 }
 
 type HealthProberType string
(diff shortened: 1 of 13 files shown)


qiujian16:
adding tests.

pkg/agent/inteface.go
```diff
@@ -117,6 +118,25 @@ type InstallStrategy struct {
 
 type HealthProber struct {
 	Type HealthProberType
+
+	WorkProber WorkHealthProber
```
yue9944882: aesthetically i'm leaning on the typical union struct:

```
type HealthProber struct {
         // Type is the type of health checker
	Type HealthProberType 
         // Work specifies how to map the status of the addon's underlying ManifestWork 
         // resource to the overall healthiness of the addon.
         // Optional if the type is not "Work"
	Work *WorkHealthProber {
             // ProbeFields is a list of resource fields to track from the ManifestWork.
             ProbeFields []ProbeField
             // FieldHealthCheckFunc is a function that asserts the healthiness of the addon.
             // Returning a nil error means the addon is healthy.
             FieldHealthCheckFunc HealthCheck(workapiv1.ResourceIdentifier, workapiv1.StatusFeedbackResult) error
         }
}         
```

pkg/utils/probe_helper.go
```diff
@@ -0,0 +1,62 @@
...
+					Type: workapiv1.WellKnownStatusType,
+				},
+			},
+		})
+	}
+	return probeFields
+}
+
+func (d *DeploymentProber) HealthCheck(identifier workapiv1.ResourceIdentifier, result workapiv1.StatusFeedbackResult) error {
+	if len(result.Values) == 0 {
```
yue9944882, line 64: returning a non-nil error means the health-check controller will retry in the default backoff strategy which has a milli-sec backoff in the first 5-ish attempts IIRC, while the default interval of refreshing the feedback values is 30s. i think 0-length feedbacks can be somewhat reasonable when the workloads are applied for the first time (it may take a short period for the controller to process), so the health controller should better retry the health-check in a more "lazy" strategy e.g. constantly a 10-sec interval. otherwise the error log will spam.

pkg/utils/probe_helper.go
```diff
@@ -0,0 +1,62 @@
...
+				{
+					Type: workapiv1.WellKnownStatusType,
+				},
+			},
+		})
+	}
+	return probeFields
+}
+
+func (d *DeploymentProber) HealthCheck(identifier workapiv1.ResourceIdentifier, result workapiv1.StatusFeedbackResult) error {
```
yue9944882: the per-resource conditions provides a status named "StatusFeedbackSynced", how do you think about checking the overall condition before digging into the fields?

qiujian16: here is a part to check the condition before checking using check func https://github.com/open-cluster-management-io/addon-framework/pull/77/files#diff-6cbce2e82f874dcc6e37cfb90ab66b42db275761808cca2abfd5c8b294d67e11R172. I will add a comment on that

pkg/utils/probe_helper.go
qiujian16: yeah, this is a good point.

pkg/agent/inteface.go
qiujian16: fixed

pkg/utils/probe_helper.go
qiujian16: we check the len of values in controller. if there is no value returned, set the condition to unknown. I think we may not need resync, since work status update will trigger another check on this.

yue9944882 (commented):
/lgtm

PR #79, opened 2022-03-01
Add an install strategy by label selector

qiujian16:

### DIFF
--- a/pkg/addonmanager/controllers/addoninstall/controller.go
+++ b/pkg/addonmanager/controllers/addoninstall/controller.go
@@ -2,13 +2,13 @@ package addoninstall
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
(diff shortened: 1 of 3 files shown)


pkg/addonmanager/controllers/addoninstall/controller.go
```diff
@@ -87,6 +87,23 @@ func (c *addonInstallController) sync(ctx context.Context, syncCtx factory.SyncC
 
 		switch addon.GetAgentAddonOptions().InstallStrategy.Type {
 		case agent.InstallAll:
+			return c.applyAddon(ctx, addonName, clusterName, addon.GetAgentAddonOptions().InstallStrategy.InstallNamespace)
+		case agent.InstallByLabel:
+			labelSelector := addon.GetAgentAddonOptions().InstallStrategy.LabelSelector
+			if labelSelector == nil {
+				return nil
```
jnpacker, line 95: Is it worth logging a warning or if there is a condition to say you've asked for the labelSelector, but you didn't provide one?

PR #81, opened 2022-03-09
should apply pre-delete manifestwork when the managedCluster is deleting

zhiweiyin318:
should apply pre-delete manifestwork when the managedCluster and managedClusterAddon are deleting.


qiujian16 (commented):
/approve
/lgtm

PR #83, opened 2022-03-10
fix Klog flags disppeared

JiahaoWei-RH:
Signed-off-by: JiaHao Wei <1335649406@qq.com>
Fix the Klog flags disppeared issue mentioned in https://github.com/open-cluster-management-io/OCM/issues/31

### DIFF
--- a/examples/helloworld/main.go
+++ b/examples/helloworld/main.go
@@ -30,8 +30,7 @@ func main() {
 	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 
-	logs.InitLogs()
-	defer logs.FlushLogs()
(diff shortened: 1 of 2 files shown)


examples/helloworld/main.go
```diff
@@ -30,6 +30,8 @@ func main() {
 	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 
+	logs.NewOptions().AddFlags(pflag.CommandLine)
+
 	logs.InitLogs()
 	defer logs.FlushLogs()
```
zhiweiyin318: I think the  `logs.InitLogs()
	defer logs.FlushLogs()` are not necessary if we use `logs.NewOptions().AddFlags(pflag.CommandLine)`

JiahaoWei-RH: done

PR #84, opened 2022-03-15
Feat: Conditional CSR controller spawning

yue9944882:


pkg/utils/csr_helpers.go
```diff
@@ -178,3 +182,28 @@ func UnionCSRApprover(approvers ...agent.CSRApproveFunc) agent.CSRApproveFunc {
 		return true
 	}
 }
+
+// IsCSRSupported checks whether the cluster supports v1 or v1beta1 csr api.
+func IsCSRSupported(nativeClient kubernetes.Interface) (bool, bool, error) {
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(nativeClient.Discovery()))
```
qiujian16, line 188: this will add additional rbac requirement for addon controller

pkg/addonmanager/manager.go
```diff
@@ -76,6 +78,11 @@ func (a *addonManager) Start(ctx context.Context) error {
 		return err
 	}
 
+	v1CSRSupported, _, err := utils.IsCSRSupported(kubeClient)
```
qiujian16, line 81: could we add an additional flag for addon to check this? the v1CSRSupported could be true initially, but the value is set if the flag for the manager is enabled. I am concerned about addition rbac requirement for addon controller.

pkg/utils/csr_helpers.go
yue9944882: the `system:discovery` clusterrole will be bound to all the authenticated users. so no additional permission is required.

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: system:discovery
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:authenticated
```

pkg/addonmanager/manager.go
yue9944882: i believe there's no worry in the RBAC compatibility. i was thinking about having a flag to manually switch between enabling/disabling CSR controllers, but by leveraging rest mapper the detection of v1 CSR api can be done completely automatically. so we dont need the flag, wdyt?

pkg/utils/csr_helpers.go
qiujian16: ah ok

qiujian16 (commented):
/approve
/lgtm

PR #85, opened 2022-03-15
Feat: Addon approver v1beta1 support

yue9944882:


qiujian16 (commented):
could we make this as a separated informer? So we only start it when v1 csr does not exist and there is only v1beta1 csr?

This could be a similar controller as csrapprove but consume v1beta1 csr. With some conversion for interface. Also we should doc the limitation.

pkg/addonmanager/controllers/certificate/csrapprove.go
```diff
@@ -127,56 +142,188 @@ func (c *csrApprovingController) sync(ctx context.Context, syncCtx factory.SyncC
...
 
 	if registrationOption.CSRApproveCheck == nil {
-		klog.V(4).Infof("addon csr %q cannont be auto approved due to approve check not defined", csr.Name)
+		klog.V(4).Infof("addon csr %q cannont be auto approved due to approve check not defined", csr.GetName())
 		return nil
 	}
 
-	approved := registrationOption.CSRApproveCheck(managedCluster, managedClusterAddon, csr)
+	var v1CSR *certificatesv1.CertificateSigningRequest
+	if assertedV1CSR, ok := csr.(*certificatesv1.CertificateSigningRequest); ok {
```
qiujian16: how about

```
switch t := csr.(type) {
  case *certificatesv1.CertificateSigningRequest:
      approved := registrationOption.CSRApproveCheck(managedCluster, managedClusterAddon, t)
       return c.approveCSRV1(ctx, t)
  case *certificatesv1beta1.CertificateSigningRequest:
      v1CSR = unsafeConvertV1beta1CSRToV1CSR(t)
      approve := registrationOption.CSRApproveCheck(managedCluster, managedClusterAddon, v1CSR)
      return c.approveCSRV1Beta1(ctx, t)
}
```

qiujian16: I think separate approve to different funcs can make it clearer

yue9944882: updated

qiujian16 (commented):
/approve
/lgtm

PR #99, opened 2022-05-27
make the chart dir support containing file path suffix

zhujian7:

Why do we need this:

In the following case, I need to set the chart dir `manifests/chart/`(with the suffix `/`) to exclude the `manifests/chart-management` files.
```
manifests
├── chart
│   ├── Chart.yaml
│   ├── templates
│   │   ├── _helpers.tpl
│   │   ├── cluster_role.yaml
│   │   ├── cluster_role_binding.yaml
│   │   ├── namespace.yaml
│   │   └── service_account.yaml
│   └── values.yaml
└── chart-management
    ├── Chart.yaml
    ├── templates
    │   ├── _helpers.tpl
    │   ├── deployment.yaml
    │   ├── role.yaml
    │   ├── role_binding.yaml
    │   ├── route.yaml
    │   ├── service.yaml
    │   └── service_account.yaml
    └── values.yaml
```

### DIFF
--- a/pkg/addonfactory/helper.go
+++ b/pkg/addonfactory/helper.go
@@ -142,7 +142,8 @@ func getFiles(manifestFS embed.FS) ([]string, error) {
 }
 
 func stripPrefix(chartPrefix, path string) string {
-	chartPrefixLen := len(strings.Split(chartPrefix, string(filepath.Separator)))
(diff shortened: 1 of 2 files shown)


qiujian16 (commented):
/approve
/lgtm

PR #100, opened 2022-06-08
Do not update addon when spec is changed

qiujian16:

### DIFF
--- a/pkg/addonmanager/controllers/addoninstall/controller.go
+++ b/pkg/addonmanager/controllers/addoninstall/controller.go
@@ -119,10 +119,11 @@ func (c *addonInstallController) sync(ctx context.Context, syncCtx factory.SyncC
 }
 
 func (c *addonInstallController) applyAddon(ctx context.Context, addonName, clusterName, installNamespace string) error {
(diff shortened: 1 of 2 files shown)


pkg/addonmanager/controllers/addoninstall/controller.go
```diff
@@ -119,10 +119,11 @@ func (c *addonInstallController) sync(ctx context.Context, syncCtx factory.SyncC
 }
 
 func (c *addonInstallController) applyAddon(ctx context.Context, addonName, clusterName, installNamespace string) error {
-	addon, err := c.managedClusterAddonLister.ManagedClusterAddOns(clusterName).Get(addonName)
-	switch {
-	case errors.IsNotFound(err):
-		addon = &addonapiv1alpha1.ManagedClusterAddOn{
+	_, err := c.managedClusterAddonLister.ManagedClusterAddOns(clusterName).Get(addonName)
```
zhiweiyin318, line 122: do we need to add the check if the managedCluster is deleting firstly?

qiujian16: It is checked upfront already

PR #102, opened 2022-06-13
WIP:hosted mode addon

zhujian7:

### DIFF
--- a/examples/helloworld/main.go
+++ b/examples/helloworld/main.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/klog/v2"
 	"open-cluster-management.io/addon-framework/examples/helloworld/agent"
 	"open-cluster-management.io/addon-framework/pkg/addonfactory"
-	addonagent "open-cluster-management.io/addon-framework/pkg/agent"
 	"open-cluster-management.io/addon-framework/pkg/version"
 
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
@@ -86,8 +85,10 @@ func runController(ctx context.Context, controllerContext *controllercmd.Control
 	agentAddon, err := addonfactory.NewAgentAddonFactory(addonName, fs, "manifests/templates").
 		WithGetValuesFuncs(getValues, addonfactory.GetValuesFromAddonAnnotation).
 		WithAgentRegistrationOption(registrationOption).
-		WithInstallStrategy(addonagent.InstallAllStrategy(agent.HelloworldAgentInstallationNamespace)).
-		BuildTemplateAgentAddon()
+		// WithInstallStrategy(addonagent.InstallAllStrategy(agent.HelloworldAgentInstallationNamespace)).
+		// BuildTemplateAgentAddon()
+		WithManagementDir("manifests/management").
+		BuildTemplateHostedAgentAddon()
 	if err != nil {
 		klog.Errorf("failed to build agent %v", err)
 		return err
--- a/pkg/addonfactory/addonfactory.go
+++ b/pkg/addonfactory/addonfactory.go
@@ -29,6 +29,7 @@ type AgentAddonFactory struct {
 	scheme            *runtime.Scheme
 	fs                embed.FS
 	dir               string
+	managementDir     string
 	getValuesFuncs    []GetValuesFunc
 	agentAddonOptions agent.AgentAddonOptions
 	// trimCRDDescription flag is used to trim the description of CRDs in manifestWork. disabled by default.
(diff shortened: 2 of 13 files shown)


pkg/utils/helpers.go
```diff
@@ -16,3 +16,20 @@ func MergeRelatedObjects(modified *bool, objs *[]addonapiv1alpha1.ObjectReferenc
 	*objs = append(*objs, obj)
 	*modified = true
 }
+
+// GetAddonInstallMode returns addon installation mode, mode could be `Default` or `Hosted`
+// TODO: Consider changing the ManagedClusterAddon API to identify the hosted mode installation
+func GetAddonInstallMode(addOn *addonapiv1alpha1.ManagedClusterAddOn) string {
+	if mode, ok := addOn.Annotations["addon.open-cluster-management.io/agent-deploy-mode"]; ok {
```
qiujian16, line 23: you could use one annotation hosting-cluster-name.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -146,65 +148,175 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
+	work, _, err := buildManifestWorkFromObject(manifestworkNamespace, managedClusterAddon.Name, objects)
 	if err != nil {
 		return err
 	}
-	work.OwnerReferences = []metav1.OwnerReference{*owner}
+
+	// FIXME: the ManagedClusterAddon can NOT be the owner as management manifests(different namespace), delete them
+	// manually when the ManagedClusterAddon is deleted.
+	if len(conditionTypePrefix) == 0 {
+		work.OwnerReferences = []metav1.OwnerReference{*owner}
```
qiujian16, line 281: this needs a finalizer on addon resource. using conditionTypePrefix to check is also tricky.

examples/helloworld/main.go
```diff
@@ -86,8 +85,10 @@ func runController(ctx context.Context, controllerContext *controllercmd.Control
 	agentAddon, err := addonfactory.NewAgentAddonFactory(addonName, fs, "manifests/templates").
 		WithGetValuesFuncs(getValues, addonfactory.GetValuesFromAddonAnnotation).
 		WithAgentRegistrationOption(registrationOption).
-		WithInstallStrategy(addonagent.InstallAllStrategy(agent.HelloworldAgentInstallationNamespace)).
-		BuildTemplateAgentAddon()
+		// WithInstallStrategy(addonagent.InstallAllStrategy(agent.HelloworldAgentInstallationNamespace)).
+		// BuildTemplateAgentAddon()
+		WithManagementDir("manifests/management").
```
qiujian16, line 90: does it mean we need to copy the same manifests in both management and hosted dir？

pkg/addonfactory/addonfactory.go
```diff
@@ -130,10 +142,52 @@ func (f *AgentAddonFactory) BuildTemplateAgentAddon() (agent.AgentAddon, error)
...
+		if err := agentAddon.validateTemplateData(file, template, "Default"); err != nil {
 			return nil, err
 		}
 		agentAddon.addTemplateData(file, template)
 	}
 	return agentAddon, nil
 }
+
+// BuildTemplateHostedAgentAddon builds a template agentAddon in Hosted mode instance.
+func (f *AgentAddonFactory) BuildTemplateHostedAgentAddon() (agent.HostedAgentAddon, error) {
```
qiujian16, line 154: why we need this?

pkg/agent/inteface.go
```diff
@@ -35,6 +35,20 @@ type AgentAddon interface {
...
+	AgentAddon
+	// ManagementManifests returns a list of manifest resources to be deployed on the management cluster for this addon.
+	// The resources in this list are required to explicitly clarify their TypeMta (i.e. apiVersion, kind)
+	// otherwise the addon deploying will fail constantly. A recommended set of addon component returned
+	// here will be:
+	//   - the hosting namespace of the addon-agents.
+	//   - a deployment of the addon agents.
+	//   - the configurations (e.g. configmaps) mounted by the deployment.
+	// NB for dispatching namespaced resources, it's recommended to include the namespace in the list.
+	ManagementManifests(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn) (
```
qiujian16, line 48: I still feel we should not have another interface. What is the benefit of doing this?

qiujian16 (commented):
I think we should be careful on the interface. I am still not clear how easily an addon can switch between these two modes. 

From the implementation perspective. I think it would be better you have a separate controller to create management manifestwork (similar as hook did). The controller should not be started by default, it should be started only when an addon saying it wants to support hosted mode.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhujian7: yes, this is just for a test and should be fixed.

examples/helloworld/main.go
zhujian7: No, only management manifests files should be held in the management dir.

zhujian7:
will use https://github.com/open-cluster-management-io/addon-framework/pull/103 to trace the hosted mode addon feature
/close

PR #111, opened 2022-07-19
refactor the hosting controller into the addondeploy controller

zhiweiyin318:


qiujian16 (commented):
this is not want i meant...

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -88,19 +95,32 @@ func NewAddonDeployController(
 				if _, ok := c.agentAddons[addonName]; !ok {
 					return false
 				}
-				if accessor.GetName() != constants.DeployWorkName(addonName) {
+				if addonNamespace, ok := accessor.GetLabels()[constants.AddonNamespaceLabel]; ok {
```
qiujian16, line 106: comments here

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -112,116 +132,221 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
 		// managed cluster is deleting, do nothing
 		return nil
 	}
 
-	managedClusterAddon, err := c.managedClusterAddonLister.ManagedClusterAddOns(clusterName).Get(addonName)
+	addon, err := c.managedClusterAddonLister.ManagedClusterAddOns(clusterName).Get(addonName)
 	if errors.IsNotFound(err) {
-		c.cache.removeCache(constants.DeployWorkName(addonName), clusterName)
-		return nil
+		return c.cleanupDeployWorks(ctx, clusterName, addonName)
```
qiujian16: shouldn't this be controlled by finalizers?

pkg/addonmanager/controllers/agentdeploy/default_sync.go
```diff
@@ -0,0 +1,98 @@
...
+			Type:    constants.AddonManifestApplied,
+			Status:  metav1.ConditionFalse,
+			Reason:  constants.AddonManifestAppliedReasonWorkApplyFailed,
+			Message: fmt.Sprintf("deleted the manifestWork %s/%s since there is no object", deployWorkNamespace, deployWorkName),
+		})
+		return addon, nil
+	}
+
+	if hookWork == nil {
+		if addonRemoveFinalizer(addon, constants.PreDeleteHookFinalizer) {
```
qiujian16: this should be hook sync, right?

qiujian16 (commented):
I still think hook should be in another sync func.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: added

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: I changed to use `c.cleanup ` that only removes cache of deploy work in default mode here.
there are 2 kinds of case:
1. there is no finalizer on the addon in default mode without pre-delete hook. I think need to remove the cache here.
2. there is finalizer on the addon in default mod with pre-delete hook and hosted mode.  cleanup will be controlled by finalizers.

pkg/addonmanager/controllers/agentdeploy/default_sync.go
zhiweiyin318: added 2 hook syncs to handle hook

pkg/addonmanager/controllers/agentdeploy/default_hook_sync.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return addon, err
+	}
+
+	if hookWork == nil {
+		addonRemoveFinalizer(addon, constants.PreDeleteHookFinalizer)
+		return addon, nil
+	}
+
+	if addonAddFinalizer(addon, constants.PreDeleteHookFinalizer) {
```
qiujian16, line 37: this is tricky are using updateStatus to update finalizer? we should update finalizer separately here.

pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		if err = s.cleanupDeployWork(ctx, addon, hostingClusterName); err != nil {
+			return addon, err
+		}
+		addonRemoveFinalizer(addon, constants.HostingManifestFinalizer)
+		return addon, nil
+	}
+
+	if addonAddFinalizer(addon, constants.HostingManifestFinalizer) {
```
qiujian16, line 92: ditto, addon/remove finalizer need to be done separately

pkg/addonmanager/controllers/agentdeploy/default_hook_sync.go
```diff
@@ -0,0 +1,73 @@
...
+		return addon, nil
+	}
+
+	// will deploy the pre-delete hook manifestWork when the addon is deleting
+	hookWork, err = s.controller.applyWork(ctx, installMode, hookWork, addon)
+	if err != nil {
+		return addon, err
+	}
+
+	if hookWorkIsCompleted(hookWork) {
```
qiujian16, line 52: add a TODO here, we should surface more message

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -112,116 +141,204 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
-	setStatusFeedbackRule(work, agentAddon)
+	return deployWork, hookWork, nil
+}
 
+func (c *addonDeployController) applyWork(ctx context.Context, installMode string,
+	work *workapiv1.ManifestWork, addon *addonapiv1alpha1.ManagedClusterAddOn) (*workapiv1.ManifestWork, error) {
+	var appliedType string
+	switch installMode {
+	case constants.InstallModeHosted:
+		appliedType = constants.AddonHostingManifestApplied
```
qiujian16: I think we should make it as an input var

pkg/addonmanager/controllers/agentdeploy/default_sync.go
```diff
@@ -0,0 +1,57 @@
...
+	if deployWork == nil {
+		err = deleteWork(ctx, s.controller.workClient, deployWorkNamespace, deployWorkName)
+		if err != nil {
+			return addon, fmt.Errorf("failed to delete work %s/%s when got no object", deployWorkNamespace, deployWorkName)
+		}
+		s.controller.cache.removeCache(deployWorkNamespace, deployWorkName)
+
+		meta.SetStatusCondition(&addon.Status.Conditions, metav1.Condition{
+			Type:    constants.AddonManifestApplied,
+			Status:  metav1.ConditionFalse,
```
qiujian16: why it is false?

pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if deployWork == nil {
+		if err = s.cleanupDeployWork(ctx, addon, hostingClusterName); err != nil {
+			return addon, fmt.Errorf("failed to cleanup deploy work %s/%s when got no object", hostingClusterName, deployWorkName)
+		}
+
+		meta.SetStatusCondition(&addon.Status.Conditions, metav1.Condition{
+			Type:    constants.AddonHostingManifestApplied,
+			Status:  metav1.ConditionFalse,
```
qiujian16: this also seems weird. we should either set it to true or do not set it.

pkg/addonmanager/controllers/agentdeploy/default_hook_sync.go
zhiweiyin318: added

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/default_sync.go
zhiweiyin318: removed the SetStatusCondition

pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
zhiweiyin318: removed the SetStatusCondition

pkg/addonmanager/controllers/agentdeploy/default_hook_sync.go
zhiweiyin318: my addon update process is:
I only add/remove finalizer to/from the addon in each sync.
in the `updateAddon` after all syncs,  check the finalizer change firstly, if has changes, update the addon and return, do not patch the condition, otherwise patch the condition.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -112,116 +141,195 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
-		if updateErr := utils.PatchAddonCondition(ctx, c.addonClient, managedClusterAddonCopy, managedClusterAddon); updateErr != nil {
-			return fmt.Errorf("failed to update managedclusteraddon status: %v; the err should be %v", updateErr, err)
+		return deployWork, hookWork, err
+	}
+	if deployWork != nil {
+		if owner != nil {
+			deployWork.OwnerReferences = []metav1.OwnerReference{*owner}
 		}
-		return err
+		setStatusFeedbackRule(deployWork, agentAddon)
```
qiujian16, line 295: hrm I think we need a comment on this option that it only supports certain resources, is it?

qiujian16 (commented):
/approve

looks good in general

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -112,116 +141,195 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
+	return err
+}
 
-	owner := metav1.NewControllerRef(managedClusterAddon, addonapiv1alpha1.GroupVersion.WithKind("ManagedClusterAddOn"))
+// cleanup is only to remove cache for deploy work in default mode when the addon is not found.
+// could append the other cleanup in the future. currently there are 2 scenarios here:
+// 1.there is no finalizer on the addon in default mode without pre-delete hook. need to remove cache here.
+// 2.there are finalizer on the addon in default mod with pre-delete hook and hosted mode. cleanup will be
+// controlled by finalizers in sync.
+func (c *addonDeployController) cleanup(ctx context.Context, clusterName, addonName string) error {
```
zhujian7: how about renaming it to cleanupCache?

pkg/addonmanager/controllers/agentdeploy/default_sync.go
```diff
@@ -0,0 +1,48 @@
...
+	if deployWork == nil {
+		err = deleteWork(ctx, s.controller.workClient, deployWorkNamespace, deployWorkName)
+		if err != nil {
+			return addon, fmt.Errorf("failed to delete work %s/%s when got no object", deployWorkNamespace, deployWorkName)
+		}
+		s.controller.cache.removeCache(deployWorkNamespace, deployWorkName)
+		return addon, nil
+	}
+
+	if !addon.DeletionTimestamp.IsZero() {
```
zhujian7: Is it better to move these before `s.controller.buildManifestWorks`?

zhujian7: and I am thinking should we also need to deleteWork(...) when `!addon.DeletionTimestamp.IsZero()`?

pkg/addonmanager/controllers/agentdeploy/default_hook_sync.go
```diff
@@ -0,0 +1,74 @@
...
+	cluster *clusterv1.ManagedCluster,
+	addon *addonapiv1alpha1.ManagedClusterAddOn) (*addonapiv1alpha1.ManagedClusterAddOn, error) {
+	installMode := constants.InstallModeDefault
+	deployWorkNamespace := addon.Namespace
+	_, hookWork, err := s.controller.buildManifestWorks(ctx, s.agentAddon, installMode, deployWorkNamespace, cluster, addon)
+	if err != nil {
+		return addon, err
+	}
+
+	if hookWork == nil {
```
zhujian7, line 32: Will this cause the finalizer to be removed and added back, and the cycle goes back and forth?

pkg/addonmanager/controllers/agentdeploy/default_sync.go
zhiweiyin318: yes. done.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: done

zhujian7:
A followup PR will make the helloworld_hosted example not depend on the helloworld_helm example. cc @zhiweiyin318 
/lgtm

PR #112, opened 2022-07-21
add an annotation to disable addon automatic installation

zhujian7:

Will not automatically install addon agent on the cluster which has `addon.open-cluster-management.io/disable-automatic-installation: "true"` annotation


zhujian7:
cc @qiujian16 @elgnay @xuezhaojun

qiujian16 (commented):
/approve
/lgtm

PR #115, opened 2022-08-02
fix cannot cleanup manifestWork when the cluster is deleting

zhiweiyin318:

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/controller.go
+++ b/pkg/addonmanager/controllers/agentdeploy/controller.go
@@ -151,20 +151,12 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 
 	cluster, err := c.managedClusterLister.Get(clusterName)
 	if errors.IsNotFound(err) {
-		// to delete the addon to make sure the addon is cleaned up when the cluster is not found.
-		// need go to sync to handle cleanup if the addon is deleting.
-		// in the normal case the addon should have been cleaned up after the cluster is deleted.
-		if addon.DeletionTimestamp.IsZero() {
-			return c.cleanupAddon(ctx, clusterName, addonName)
-		}
+		// the managedCluster is nil in this case,and sync cannot handle nil managedCluster.
+		// TODO: consider to force delete the addon and its deploy manifestWorks.
+		return nil
 	}
 	if err != nil {
 		return err
-	}
-
-	if !cluster.DeletionTimestamp.IsZero() {
-		// TODO: consider to delete addon here
-		return nil
 	}
 
 	syncers := []addonDeploySyncer{
@@ -200,14 +192,6 @@ func (c *addonDeployController) updateAddon(ctx context.Context, new, old *addon
 	}
 
 	return utils.PatchAddonCondition(ctx, c.addonClient, new, old)
-}
-
-func (c *addonDeployController) cleanupAddon(ctx context.Context, clusterName, addonName string) error {
-	err := c.addonClient.AddonV1alpha1().ManagedClusterAddOns(clusterName).Delete(ctx, addonName, metav1.DeleteOptions{})
-	if errors.IsNotFound(err) {
-		return nil
-	}
-	return err
 }
 
 // cleanupCache is only to remove cache for deploy work in default mode when the addon is not found.
--- a/pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
+++ b/pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
@@ -93,6 +93,12 @@ func (s *hostedSyncer) sync(ctx context.Context,
 		return addon, nil
 	}
 
+	// waiting for the addon to be deleted when cluster is deleting.
+	// TODO: consider to delete addon in this scenario.
+	if !cluster.DeletionTimestamp.IsZero() {
+		return addon, nil
+	}
+
 	deployWorkName := constants.DeployHostingWorkName(addon.Namespace, addon.Name)
 	deployWork, _, err := s.controller.buildManifestWorks(ctx, s.agentAddon, installMode, hostingClusterName, cluster, addon)
 	if err != nil {
--- a/pkg/addonmanager/controllers/agentdeploy/default_sync.go
+++ b/pkg/addonmanager/controllers/agentdeploy/default_sync.go
@@ -29,6 +29,12 @@ func (s *defaultSyncer) sync(ctx context.Context,
 		return addon, nil
 	}
 
+	// waiting for the addon to be deleted when cluster is deleting.
+	// TODO: consider to delete addon in this scenario.
+	if !cluster.DeletionTimestamp.IsZero() {
+		return addon, nil
+	}
+
 	deployWork, _, err := s.controller.buildManifestWorks(ctx, s.agentAddon, installMode, deployWorkNamespace, cluster, addon)
 	if err != nil {
 		return addon, err
--- a/pkg/addonmanager/controllers/agentdeploy/default_sync_test.go
+++ b/pkg/addonmanager/controllers/agentdeploy/default_sync_test.go
@@ -53,15 +53,13 @@ func TestDefaultReconcile(t *testing.T) {
 		validateWorkActions  func(t *testing.T, actions []clienttesting.Action)
 	}{
 		{
-			name:         "no cluster",
-			key:          "cluster1/test",
-			addon:        []runtime.Object{addontesting.NewAddon("test", "cluster1")},
-			cluster:      []runtime.Object{},
-			existingWork: []runtime.Object{},
-			validateAddonActions: func(t *testing.T, actions []clienttesting.Action) {
-				addontesting.AssertActions(t, actions, "delete")
-			},
-			validateWorkActions: addontesting.AssertNoActions,
+			name:                 "no cluster",
+			key:                  "cluster1/test",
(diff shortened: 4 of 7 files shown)


pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -150,20 +150,21 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
-		}
+
+	// waiting for the addon is deleted after cluster is deleted.
+	// need go to sync to handle cleanup if the addon is deleting.
+	// The other controller should delete the addon once cluster is deleted.
+	// TODO: should consider to force delete the addon and its manifestWork because there may be no work-agent in spoke cluster.
+	if errors.IsNotFound(err) && addon.DeletionTimestamp.IsZero() {
+		return nil
 	}
 	if err != nil {
```
zhujian7: if the err is not found err and addon is in deletion, will it return not found error here?

zhujian7:
is it possible to add a ut or other test?

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: yes.   fixed to return when meet err. aded a TODO here.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -28,6 +28,7 @@ import (
 	worklister "open-cluster-management.io/api/client/work/listers/work/v1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	workapiv1 "open-cluster-management.io/api/work/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
qiujian16: why use this? I would not use this for not found error

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -150,20 +151,14 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
-		// need go to sync to handle cleanup if the addon is deleting.
-		// in the normal case the addon should have been cleaned up after the cluster is deleted.
-		if addon.DeletionTimestamp.IsZero() {
-			return c.cleanupAddon(ctx, clusterName, addonName)
-		}
-	}
 	if err != nil {
-		return err
+		// TODO: should consider to force delete the addon and its manifestWork if cluster is not found.
+		return client.IgnoreNotFound(err)
```
qiujian16: I do not understand, if the cluster is not found, shouldn't we remove the work?

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -150,20 +151,14 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
 	}
 
-	if !cluster.DeletionTimestamp.IsZero() {
-		// TODO: consider to delete addon here
+	// waiting for the addon is deleted after cluster is deleted.
+	// need go to sync to handle cleanup if the addon is deleting.
+	if !cluster.DeletionTimestamp.IsZero() && addon.DeletionTimestamp.IsZero() {
```
qiujian16: this is also confusing, why you need to care addon status.

Or you should care about cluster status in each syncer instead?

qiujian16:
please add an integration test for this.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: removed

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: the cluster will be nil if get `not found` here, so cannot go into sync with a nil cluster because there are many functions using cluster without checking if cluster is valid or not.
add a TODO here to force delete the addon and manifestWork if the cluster is not found when we begin to handle cluster deletion in the future.

pkg/addonmanager/controllers/agentdeploy/controller.go
zhiweiyin318: moved the cluster status check to each syncer.

when the cluster is deleting, there are 2 cases:
1. the addon is not deleting, we should not apply/update the deploy manifestWork, and wait until the addon is deleted.
2. the addon is deleting, we begin the cleanup process to remove the deploy manifestWork.

zhiweiyin318:
> please add an integration test for this.

added

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -150,20 +150,12 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
...
-	if errors.IsNotFound(err) {
-		// to delete the addon to make sure the addon is cleaned up when the cluster is not found.
-		// need go to sync to handle cleanup if the addon is deleting.
-		// in the normal case the addon should have been cleaned up after the cluster is deleted.
-		if addon.DeletionTimestamp.IsZero() {
-			return c.cleanupAddon(ctx, clusterName, addonName)
-		}
-	}
-	if err != nil {
+	if err != nil && !errors.IsNotFound(err) {
```
zhujian7:
```suggestion
	if errors.IsNotFound(err) {
		// the managedCluster is nil in this case,and sync cannot handle nil managedCluster.
		// TODO: consider to force delete the addon and its deploy manifestWorks.
		return nil
	}
	if err != nil {
		return err
	}
```

test/integration/agent_hosting_deploy_test.go
```diff
@@ -217,6 +217,26 @@ var _ = ginkgo.Describe("Agent deploy", func() {
...
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		err = hubClusterClient.ClusterV1().ManagedClusters().Delete(context.Background(), managedClusterName, metav1.DeleteOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		gomega.Eventually(func() error {
+			time.Sleep(5 * time.Second) // wait 5 seconds to sync
+			work, err = hubWorkClient.WorkV1().ManifestWorks(hostingClusterName).Get(context.Background(),
+				hostingManifestWorkName, metav1.GetOptions{})
+			if err != nil {
+				gomega.Expect(errors.IsNotFound(err)).ToNot(gomega.BeTrue())
+				return err
```
zhujian7: I don't quite understand here, will it return err even if it is a not found err?

test/integration/agent_hosting_deploy_test.go
```diff
@@ -217,6 +217,26 @@ var _ = ginkgo.Describe("Agent deploy", func() {
 			return nil
 		}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
 
+		// do nothing if cluster is deleting and addon is not deleted
+		cluster, err := hubClusterClient.ClusterV1().ManagedClusters().Get(context.Background(), managedClusterName, metav1.GetOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		cluster.SetFinalizers([]string{"cluster.open-cluster-management.io/api-resource-cleanup"})
```
zhujian7, line 223: does this finalizer aim at preventing the cluster be deleted？Is it necessary to remove it after the test is finished?

test/integration/agent_hosting_deploy_test.go
zhiweiyin318: fixed

test/integration/agent_hosting_deploy_test.go
zhiweiyin318: this is integration test, not necessary to cleanup.

pkg/addonmanager/controllers/agentdeploy/controller.go
qiujian16: but we do remove manifestwork when cluster is not found, shouldn't this change break the previous behavior?

zhiweiyin318: we only return nil when cluster is not found  before my refactor PR.  I deleted addon but not manifestWork when cluster is not found in my refactor PR.  
This is an exception scenario, so I think only to delete addon is not a good option, we should force deleted addon and manifestWork in this scenario, should we do the force deletion in addon-framework or other component like deletion controller of registration?

pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
```diff
@@ -93,6 +93,12 @@ func (s *hostedSyncer) sync(ctx context.Context,
 		return addon, nil
 	}
 
+	// waiting for the addon to be deleted when cluster is deleting.
+	// TODO: consider to delete addon in this scenario.
+	if !cluster.DeletionTimestamp.IsZero() {
```
qiujian16, line 98: so the order of this line of code matters, we must check this after checking addon, right?

A better way is we have a finalize sync separately, which handles those finalize things.

qiujian16:
I can understand the intention here, but I still think it could be error prone. We could merge it if it is urgent. But we should have a followup to refactor this part.

/approve
/lgtm
/hold

zhujian7:
> I can understand the intention here, but I still think it could be error prone. We could merge it if it is urgent. But we should have a followup to refactor this part.

yeah, I agree it is really error prone, we need to refactor it. but it is a little bit urgent I will unhold it currently.
/unhold

PR #117, opened 2022-08-16
Using index in agent deploy

qiujian16:
This is for splitting work to multiple

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/default_sync.go
+++ b/pkg/addonmanager/controllers/agentdeploy/default_sync.go
@@ -2,17 +2,25 @@ package agentdeploy
 
 import (
 	"context"
-	"fmt"
 
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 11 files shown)


pkg/addonmanager/controllers/agentdeploy/default_sync.go
```diff
@@ -35,21 +53,32 @@ func (s *defaultSyncer) sync(ctx context.Context,
...
-	deployWork, _, err := s.controller.buildManifestWorks(ctx, s.agentAddon, installMode, deployWorkNamespace, cluster, addon)
+	deployWorks, _, err := buildManifestWorks(ctx, s.agentAddon, installMode, deployWorkNamespace, cluster, addon)
 	if err != nil {
 		return addon, err
 	}
 
-	// deployWork is nil only in the case that there is no object from addon manifests. need to delete the deployWork.
-	if deployWork == nil {
-		err = deleteWork(ctx, s.controller.workClient, deployWorkNamespace, deployWorkName)
+	requiredWorkNames := sets.NewString()
```
zhujian7, line 56: I noticed that the logic of comparing current work and required work and updating is almost the same as hosted_sync, is it possible to extract a common function like updateWorks?

PR #122, opened 2022-09-15
add work builder to build and apply multi works

zhiweiyin318:

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/utils.go
+++ b/pkg/addonmanager/controllers/agentdeploy/utils.go
@@ -1,12 +1,16 @@
 package agentdeploy
 
 import (
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	"open-cluster-management.io/addon-framework/pkg/addonmanager/constants"
+	"open-cluster-management.io/addon-framework/pkg/agent"
+	"open-cluster-management.io/addon-framework/pkg/common/workbuilder"
 	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
 	workapiv1 "open-cluster-management.io/api/work/v1"
 )
@@ -83,7 +87,7 @@ func newManifestWork(addonNamespace, addonName, clusterName string, manifests []
 // isPreDeleteHookObject check the object is a pre-delete hook resources.
 // currently, we only support job and pod as hook resources.
 // we use WellKnownStatus here to get the job/pad status fields to check if the job/pod is completed.
-func (b *manifestWorkBuiler) isPreDeleteHookObject(obj runtime.Object) (bool, *workapiv1.ManifestConfigOption) {
+func (b *addonWorksBuilder) isPreDeleteHookObject(obj runtime.Object) (bool, *workapiv1.ManifestConfigOption) {
 	var resource string
 	gvk := obj.GetObjectKind().GroupVersionKind()
 	switch gvk.Kind {
@@ -118,28 +122,31 @@ func (b *manifestWorkBuiler) isPreDeleteHookObject(obj runtime.Object) (bool, *w
 		},
 	}
 }
-func newManagedManifestWorkBuilder(hostedModeEnabled bool) *manifestWorkBuiler {
-	return &manifestWorkBuiler{
+func newAddonWorksBuilder(hostedModeEnabled bool, workBuilder *workbuilder.WorkBuilder) *addonWorksBuilder {
+	return &addonWorksBuilder{
 		processor:         &managedManifest{},
 		hostedModeEnabled: hostedModeEnabled,
+		workBuilder:       workBuilder,
 	}
 }
 
-func newHostingManifestWorkBuilder(hostedModeEnabled bool) *manifestWorkBuiler {
-	return &manifestWorkBuiler{
+func newHostingAddonWorksBuilder(hostedModeEnabled bool, workBuilder *workbuilder.WorkBuilder) *addonWorksBuilder {
+	return &addonWorksBuilder{
 		processor:         &hostingManifest{},
 		hostedModeEnabled: hostedModeEnabled,
+		workBuilder:       workBuilder,
 	}
 }
 
-type manifestWorkBuiler struct {
+type addonWorksBuilder struct {
 	processor         manifestProcessor
 	hostedModeEnabled bool
+	workBuilder       *workbuilder.WorkBuilder
 }
 
 type manifestProcessor interface {
 	deployable(hostedModeEnabled bool, installMode string, obj runtime.Object) (bool, error)
-	manifestWorkName(addonNamespace, addonName string) string
+	manifestWorkNamePrefix(addonNamespace, addonName string) string
 	preDeleteHookManifestWorkName(addonNamespace, addonName string) string
 }
 
@@ -175,8 +182,8 @@ func (m *hostingManifest) deployable(hostedModeEnabled bool, installMode string,
 	return false, nil
 }
 
-func (m *hostingManifest) manifestWorkName(addonNamespace, addonName string) string {
-	return constants.DeployHostingWorkName(addonNamespace, addonName)
+func (m *hostingManifest) manifestWorkNamePrefix(addonNamespace, addonName string) string {
+	return constants.DeployHostingWorkNamePrefix(addonNamespace, addonName)
 }
 
 func (m *hostingManifest) preDeleteHookManifestWorkName(addonNamespace, addonName string) string {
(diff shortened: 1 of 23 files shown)


pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+	namespace, name string
+}
+
+type requiredManifestMapper map[manifestKey]workapiv1.Manifest
+
+// GetExistingManifestWorks is to get existing manifestWorks which need to be updated.
+type GetExistingManifestWorks func() ([]workapiv1.ManifestWork, error)
+
+// GenerateManifestWork is to generate a new manifestWork with empty manifest, need to make sure the name is unique.
+type GenerateManifestWork func(index int) workapiv1.ManifestWork
```
qiujian16: GenerateManifestWorkObjectMeta and only return metav1.ObjectMeta

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		return key, err
+	}
+	key.namespace = accessor.GetNamespace()
+	key.name = accessor.GetName()
+	return key, nil
+}
+
+func buildManifest(object runtime.Object) (workapiv1.Manifest, error) {
+	rawObject, err := runtime.Encode(unstructured.UnstructuredJSONScheme, object)
```
qiujian16: are you sure it works? what if the object is not unstructure?

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+
+// WithWorkApplier is to set the workApplier used to apply the manifestWorks.
+// This is a required setting for BuildAndApply.
+func (w *WorkBuilder) WithWorkApplier(workApplier *workapplier.WorkApplier) *WorkBuilder {
+	w.workApplier = workApplier
+	return w
+}
+
+// NewBuild is to build a set of manifestWorks using the objects.
+func (w *WorkBuilder) NewBuild(objects []runtime.Object) ([]workapiv1.ManifestWork, error) {
```
qiujian16: BuildNew, add a comment that it is not to consider existing work

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+
+type manifestKey struct {
+	gvk             schema.GroupVersionKind
+	namespace, name string
+}
+
+type requiredManifestMapper map[manifestKey]workapiv1.Manifest
+
+// GetExistingManifestWorks is to get existing manifestWorks which need to be updated.
+type GetExistingManifestWorks func() ([]workapiv1.ManifestWork, error)
```
qiujian16: I am not sure how it works, you need an input as key, right?

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+
+func generateRequiredManifestMapper(objects []runtime.Object) (requiredManifestMapper, error) {
+	var mapper = requiredManifestMapper{}
+	for _, object := range objects {
+		rawObject, err := runtime.Encode(unstructured.UnstructuredJSONScheme, object)
+		if err != nil {
+			return nil, fmt.Errorf("failed to encode object %v, err: %v", object, err)
+		}
+		key := manifestKey{
+			gvk: object.GetObjectKind().GroupVersionKind(),
```
qiujian16: you might need to return err when kind or version is not set

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+		return workapiv1.Manifest{}, fmt.Errorf("failed to encode object %v, err: %v", object, err)
+	}
+	manifest := workapiv1.Manifest{RawExtension: runtime.RawExtension{Raw: rawObject}}
+	return manifest, nil
+}
+
+func generateRequiredManifestMapper(objects []runtime.Object) (requiredManifestMapper, error) {
+	var mapper = requiredManifestMapper{}
+	for _, object := range objects {
+		rawObject, err := runtime.Encode(unstructured.UnstructuredJSONScheme, object)
```
qiujian16, line 336: what if object is not unstructured.

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+			return nil, err
+		}
+		key.namespace = accessor.GetNamespace()
+		key.name = accessor.GetName()
+		mapper[key] = workapiv1.Manifest{RawExtension: runtime.RawExtension{Raw: rawObject}}
+	}
+	return mapper, nil
+}
+
+func (w *WorkBuilder) compareManifests(existingWorks []workapiv1.ManifestWork,
```
qiujian16: maybe syncManifests

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+	requiredMapper requiredManifestMapper) (appliedWorks, deletedWorks []workapiv1.ManifestWork, err error) {
+	var updatedWorks []workapiv1.ManifestWork
+	for workIndex := 0; workIndex < len(existingWorks); workIndex++ {
```
qiujian16: add a comment here, this step is to update existing manifestwork, delete the non-existing manifests, update existing manifets.

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+		sort.SliceStable(updatedWorks, func(i, j int) bool {
+			iTotalSize := totalSizeOfManifests(updatedWorks[i].Spec.Workload.Manifests)
+			jTotalSize := totalSizeOfManifests(updatedWorks[j].Spec.Workload.Manifests)
+			return jTotalSize > iTotalSize
+		})
+
+		var newWorkManifests []workapiv1.Manifest
+		for manifestIndex := 0; manifestIndex < len(createdManifests); manifestIndex++ {
+			appended := false
+			for workIndex := 0; workIndex < len(updatedWorks); workIndex++ {
```
qiujian16: it does not look correct. I think you can sort the createdManifest at first

In the for loop, you would also need to find the min size existing work,

if the min size cannot occupy, go to create one and added into the existing work.

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	if len(manifests) == 0 {
+		return nil, fmt.Errorf("there is no manifest")
+	}
+
+	return w.newManifestWorks(manifests, 0)
+}
+
+// BuildAndApply is to build a set manifestWorks using the objects and update the existing manifestWorks.
+func (w *WorkBuilder) BuildAndApply(ctx context.Context, objects []runtime.Object) ([]workapiv1.ManifestWork, error) {
```
qiujian16: func (w *WorkBuilder) BuildAndApply(ctx context.Context, objects []runtime.Object, opts....) {
   builder := internalBuilder()
   for _, opts := range opts {
      builder = opts(builder)
  }

  appliedWorks, deletedWorks, err := builder.buildManifestWorks(objects)
}

ExistingWorkOptions(existingWork ) *BuilderOption {
   return func(builder InternalBuilder) InternalBuilder {
     builder.existingWork = existingWork
     return builder
} 

DeleteOption()

FeedBackOption()

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: done

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: verified it works.

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: changed to 
```
func (w *WorkBuilder) Build(objects []runtime.Object,
	generateManifestWorkObjectMeta GenerateManifestWorkObjectMeta,
	options ...WorkBuilderOption) (applied, deleted []workapiv1.ManifestWork, err error) {
```

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: deleted instead of  ExistingManifestWorksOption

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: done

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: verified it works.

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: done

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: done

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: fixed.

pkg/common/workbuilder/workbuilder.go
zhiweiyin318: done

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -216,63 +223,100 @@ func (m *managedManifest) deployable(hostedModeEnabled bool, installMode string,
...
-		hookManifestWork.Spec.ManifestConfigs = manifestConfigs
+		hookManifests = append(hookManifests, workapiv1.Manifest{RawExtension: runtime.RawExtension{Raw: rawObject}})
+		hookManifestConfigs = append(hookManifestConfigs, *manifestConfig)
+	}
+	if len(hookManifests) == 0 {
+		return nil, nil
 	}
 
-	return deployManifestWork, hookManifestWork, nil
+	hookWork = newManifestWork(addon.Namespace, addon.Name, addonWorkNamespace, hookManifests, b.processor.preDeleteHookManifestWorkName)
```
qiujian16, line 315: can we use the same build func?

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -216,63 +223,100 @@ func (m *managedManifest) deployable(hostedModeEnabled bool, installMode string,
...
-	var manifestConfigs []workapiv1.ManifestConfigOption
-
-	addonName := addon.Name
+	existingWorks []workapiv1.ManifestWork,
+	objects []runtime.Object,
+	manifestOptions []workapiv1.ManifestConfigOption) (deployWorks, deleteWorks []*workapiv1.ManifestWork, err error) {
+	var deployObjects []runtime.Object
+	var owner *metav1.OwnerReference
 	installMode, _ := constants.GetHostedModeInfo(addon.GetAnnotations())
+	if installMode == constants.InstallModeDefault {
```
qiujian16, line 246: add a comment here, hosted mode should not have ownerref.

qiujian16:
/approve

pass to @zhujian7 

Also I am wondering whether we still need another level of addonWorkBuilder, or we can just use buildAndApply? But this could be a followup.

pkg/common/workbuilder/workbuilder.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+type manifestSize struct {
+	manifest workapiv1.Manifest
+	size     int
+}
+
+type manifestWorkBuffer struct {
+	work   *workapiv1.ManifestWork
+	buffer int
```
zhujian7, line 38: could you add some comments for the field `buffer`?

zhiweiyin318: added

pkg/addonmanager/controllers/agentdeploy/utils.go
zhiweiyin318: added

pkg/addonmanager/controllers/agentdeploy/utils.go
zhiweiyin318: yes, but it will not be easy to understand the returned work slices since we only define only a manifestWork for the pre-delete hook manifests, and do not need existing works to build.

PR #126, opened 2022-09-29
Allow trigger reconcile externally

### LINKED ISSUE #96: Allow triggering addon reconciliation
I am trying to create an addon for OCM, the goal is to deploy workload based on an external configuration.

I managed to follow the path of the "hello world" addon, deploying the "hello-world" configmap into the managed/target cluster.

However, I would like to have more dynamic workload deployed to the managed cluster. To my understanding, whatever the method `Manifest` of the addon returns, will be deployed on the managed cluster.

I only can't find a way, how to re-trigger this reconcile loop. So that a change in our system triggers the `Manfiest` function to be evaluated again, so that the new state will will rendered.

qiujian16:

fixes https://github.com/open-cluster-management-io/addon-framework/issues/96

### DIFF
--- a/pkg/addonmanager/manager.go
+++ b/pkg/addonmanager/manager.go
@@ -37,6 +37,10 @@ type AddonManager interface {
 	// AddAgent register an addon agent to the manager.
 	AddAgent(addon agent.AgentAddon) error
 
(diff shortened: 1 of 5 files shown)


skeeey:
LGTM

PR #135, opened 2022-11-30
Remove unsupported config condition if config is corrected

skeeey:
Signed-off-by: Wei Liu <liuweixa@redhat.com>

### DIFF
--- a/pkg/addonmanager/controllers/clustermanagement/controller.go
+++ b/pkg/addonmanager/controllers/clustermanagement/controller.go
@@ -126,6 +126,18 @@ func (c *clusterManagementController) sync(ctx context.Context, syncCtx factory.
 		return utils.PatchAddonCondition(ctx, c.addonClient, addonCopy, addon)
 	}
 
+	// Update unsupported configuration condition if configuration becomes corrected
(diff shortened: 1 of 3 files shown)


pkg/addonmanager/controllers/clustermanagement/controller.go
```diff
@@ -126,6 +126,13 @@ func (c *clusterManagementController) sync(ctx context.Context, syncCtx factory.
 		return utils.PatchAddonCondition(ctx, c.addonClient, addonCopy, addon)
 	}
 
+	// Remove unsupported configuration condition if configuration becomes corrected
+	unsupportedConfigCondition := meta.FindStatusCondition(addon.Status.Conditions, UnsupportedConfigurationType)
+	if unsupportedConfigCondition != nil {
+		meta.RemoveStatusCondition(&addonCopy.Status.Conditions, UnsupportedConfigurationType)
```
qiujian16: why we remove this condition instead of setting it to false?

skeeey: updated, it better to set it to false than removing it

qiujian16 (commented):
/approve
/lgtm

PR #136, opened 2022-12-02
Ignore empty supported configs check if no configs in the mca

skeeey:
Signed-off-by: Wei Liu <liuweixa@redhat.com>


qiujian16 (commented):
/approve
/lgtm

PR #139, opened 2022-12-06
fix the manifestWork in hosted cluster ns is not deleted in hosted mode

zhiweiyin318:

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/utils.go
+++ b/pkg/addonmanager/controllers/agentdeploy/utils.go
@@ -242,10 +242,10 @@ func (b *addonWorksBuilder) BuildDeployWorks(addonWorkNamespace string,
 	var owner *metav1.OwnerReference
 	installMode, _ := constants.GetHostedModeInfo(addon.GetAnnotations())
 
-	// only set addon as the owner of works in default mode. should not set owner in hosted mode.
-	if installMode == constants.InstallModeDefault {
-		owner = metav1.NewControllerRef(addon, addonapiv1alpha1.GroupVersion.WithKind("ManagedClusterAddOn"))
-	}
+	// This owner is only added to the manifestWork deployed in managed cluster ns.
+	// the manifestWork in managed cluster ns is cleaned up via the addon ownerRef, so need to add the owner.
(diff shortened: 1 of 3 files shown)


pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -242,10 +242,8 @@ func (b *addonWorksBuilder) BuildDeployWorks(addonWorkNamespace string,
 	var owner *metav1.OwnerReference
 	installMode, _ := constants.GetHostedModeInfo(addon.GetAnnotations())
 
-	// only set addon as the owner of works in default mode. should not set owner in hosted mode.
-	if installMode == constants.InstallModeDefault {
-		owner = metav1.NewControllerRef(addon, addonapiv1alpha1.GroupVersion.WithKind("ManagedClusterAddOn"))
-	}
+	// The manifestWork deployed in cluster ns should have the owner in hosted mode.
+	owner = metav1.NewControllerRef(addon, addonapiv1alpha1.GroupVersion.WithKind("ManagedClusterAddOn"))
```
qiujian16, line 248: will it owner cross namespace?

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -427,7 +425,7 @@ func newAddonWorkObjectMeta(namePrefix, addonName, addonNamespace, workNamespace
 				constants.AddonLabel: addonName,
 			},
 		}
-		if owner != nil {
+		if addonNamespace == workNamespace && owner != nil {
```
qiujian16, line 433: I see what is going on here, but we need to doc this.

zhiweiyin318: added doc here . the owner is only added to manifestWork deployed in the managed cluster ns.

pkg/addonmanager/controllers/agentdeploy/utils.go
zhiweiyin318: added doc here . the owner is only added to manifestWork deployed in the managed cluster ns.

qiujian16 (commented):
/approve
/lgtm

PR #141, opened 2022-12-16
Bump to controller-runtime 0.14.1

skitt:
This is required to bump the Kubernetes dependencies to 0.26. The Informer AddEventHandler function now returns a
ResourceEventHandlerRegistration.

Signed-off-by: Stephen Kitt <skitt@redhat.com>


pkg/basecontroller/factory/factory.go
```diff
@@ -31,7 +31,7 @@ type Factory struct {
 // Informer represents any structure that allow to register event handlers and informs if caches are synced.
 // Any SharedInformer will comply.
 type Informer interface {
-	AddEventHandler(handler cache.ResourceEventHandler)
+	AddEventHandler(handler cache.ResourceEventHandler) (cache.ResourceEventHandlerRegistration, error)
```
qiujian16, line 35: I think this will not work with sharedinformer. Maybe we should have another interface that just for informer from controller-runtime

skitt: @qiujian16 `SharedInformer` has been updated too, see https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L143

skitt:
@qiujian16 we need this for Submariner in the upcoming ACM release, I don’t think there are any blockers on the OCM side; could you review this again please?

PR #149, opened 2023-02-13
Add the Helm default variable of hostingClusterCapabilities

mprahl:
When deploying in hosted mode, the managed cluster's Kubernetes version
is always used in Helm charts conditionals such as:
{{- if semverCompare "< 1.16.0" .Capabilities.KubeVersion.Version }}

The hosted mode's version is required to guard against using Kubernetes API
features that aren't present on the cluster that the addon is being
deployed to not necessarily based on the target cluster that addon will
manage.

The new default Helm chart variable of hostingClusterCapabilities
is added so that in hosted mode, the hosted cluster's Kubernetes
version Kubernetes version is available. Unless the AgentAddonFactory
was built with the WithHostingCluster method, the
hostingClusterCapabilities will have empty values in it. The empty
values being returned by default make it easier to make if statements in
the Helm chart rather than guarding against a value not being present.

Relates:
https://issues.redhat.com/browse/ACM-3233
https://issues.redhat.com/browse/ACM-2923

### DIFF
--- a/pkg/addonfactory/addonfactory.go
+++ b/pkg/addonfactory/addonfactory.go
@@ -34,6 +34,7 @@ type AgentAddonFactory struct {
 	agentAddonOptions agent.AgentAddonOptions
 	// trimCRDDescription flag is used to trim the description of CRDs in manifestWork. disabled by default.
 	trimCRDDescription bool
+	hostingCluster     *clusterv1.ManagedCluster
 }
 
 // NewAgentAddonFactory builds an addonAgentFactory instance with addon name and fs.
@@ -112,6 +113,13 @@ func (f *AgentAddonFactory) WithTrimCRDDescription() *AgentAddonFactory {
 // WithConfigGVRs defines the addon supported configuration GroupVersionResource
 func (f *AgentAddonFactory) WithConfigGVRs(gvrs ...schema.GroupVersionResource) *AgentAddonFactory {
 	f.agentAddonOptions.SupportedConfigGVRs = append(f.agentAddonOptions.SupportedConfigGVRs, gvrs...)
+	return f
+}
+
+// WithHostingCluster defines the hosting cluster used in hosted mode. An AgentAddon may use this to provide
+// additional metadata.
+func (f *AgentAddonFactory) WithHostingCluster(cluster *clusterv1.ManagedCluster) *AgentAddonFactory {
+	f.hostingCluster = cluster
 	return f
 }
 
--- a/pkg/addonfactory/helm_agentaddon.go
+++ b/pkg/addonfactory/helm_agentaddon.go
@@ -35,8 +35,9 @@ type helmBuiltinValues struct {
 // the values in helm chart should begin with a lowercase letter, so we need convert it to Values by JsonStructToValues.
 // the default values can be overrided by getValuesFuncs
 type helmDefaultValues struct {
-	HubKubeConfigSecret     string `json:"hubKubeConfigSecret,omitempty"`
-	ManagedKubeConfigSecret string `json:"managedKubeConfigSecret,omitempty"`
+	HubKubeConfigSecret        string                 `json:"hubKubeConfigSecret,omitempty"`
+	ManagedKubeConfigSecret    string                 `json:"managedKubeConfigSecret,omitempty"`
+	HostingClusterCapabilities chartutil.Capabilities `json:"hostingClusterCapabilities,omitempty"`
 }
 
 type HelmAgentAddon struct {
@@ -45,6 +46,7 @@ type HelmAgentAddon struct {
 	getValuesFuncs     []GetValuesFunc
 	agentAddonOptions  agent.AgentAddonOptions
 	trimCRDDescription bool
+	hostingCluster     *clusterv1.ManagedCluster
 }
 
 func newHelmAgentAddon(factory *AgentAddonFactory, chart *chart.Chart) *HelmAgentAddon {
@@ -54,6 +56,7 @@ func newHelmAgentAddon(factory *AgentAddonFactory, chart *chart.Chart) *HelmAgen
 		getValuesFuncs:     factory.getValuesFuncs,
 		agentAddonOptions:  factory.agentAddonOptions,
 		trimCRDDescription: factory.trimCRDDescription,
+		hostingCluster:     factory.hostingCluster,
 	}
 }
 
@@ -214,6 +217,10 @@ func (a *HelmAgentAddon) getDefaultValues(
 	}
 
 	defaultValues.ManagedKubeConfigSecret = fmt.Sprintf("%s-managed-kubeconfig", addon.Name)
+
+	if a.hostingCluster != nil {
+		defaultValues.HostingClusterCapabilities = *a.capabilities(a.hostingCluster, addon)
+	}
 
 	helmDefaultValues, err := JsonStructToValues(defaultValues)
 	if err != nil {
--- a/pkg/addonfactory/helm_agentaddon_test.go
+++ b/pkg/addonfactory/helm_agentaddon_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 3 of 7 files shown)


pkg/addonfactory/helm_agentaddon.go
```diff
@@ -171,8 +175,14 @@ func (a *HelmAgentAddon) getValues(
 
 	overrideValues = MergeValues(overrideValues, builtinValues)
 
-	values, err := chartutil.ToRenderValues(a.chart, overrideValues,
-		a.releaseOptions(cluster, addon), a.capabilities(cluster, addon))
+	var capabilities *chartutil.Capabilities
+	if hostingCluster == nil {
```
qiujian16: I think we should have another variable to define version rather than the kubeVersion, you may need to check both kube version on mc and hosted cluster. Since manifests could be deployed on both clusters.

pkg/agent/inteface.go
```diff
@@ -31,7 +31,8 @@ type AgentAddon interface {
 	//   - the RBAC permission bond to the addon agents *in the managed cluster*. (the hub cluster's RBAC
 	//     setup shall be done at GetAgentAddonOptions below.)
 	// NB for dispatching namespaced resources, it's recommended to include the namespace in the list.
-	Manifests(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn) ([]runtime.Object, error)
+	// Note that hostingCluster should be set to nil when not deploying in hosted mode.
+	Manifests(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn, hostingCluster *clusterv1.ManagedCluster) ([]runtime.Object, error)
```
qiujian16: this is a breaking change. I would rather not directly add a hostingCluster var since it is not required for all addons. Probably you could have a ops var here and allow different options.

mprahl:
@qiujian16 I took another approach. Could you please check the latest push?

zhiweiyin318:
@mprahl  @qiujian16  I have another  option for discussion. 
we can get the kube version of management cluster in `getValue` when build the agent.AgentAddon.  This has minimal impact.  https://github.com/stolostron/governance-policy-addon-controller/blob/d465196ab6d9ec6787ccd9e8643e69396928d2ba/pkg/addon/certpolicy/agent_addon.go#L37
```go
func NewGetValuesFunc(clusterClient  clusterclientset.Interface) addonfactory.GetValuesFunc {
	return func(cluster *clusterv1.ManagedCluster,
		addon *addonapiv1alpha1.ManagedClusterAddOn) (addonfactory.Values, error) {
		userValues := policyaddon.UserValues{
			GlobalValues: policyaddon.GlobalValues{
				managementClusterKubeVersion string
				...
		}
		if cluster is hosting mode {
			managementCluster,err:= clusterClient.ClusterV1().ManagedClusters().Get(managementCluster)
			userValues.GlobalValues.managementClusterKubeVersion = managementCluster.Status.Version.Kubernetes
		}

		return addonfactory.JsonStructToValues(userValues)
	}
}
```

mprahl:
> @mprahl @qiujian16 I have another option for discussion. we can get the kube version of management cluster in `getValue` when build the agent.AgentAddon. This has minimal impact. https://github.com/stolostron/governance-policy-addon-controller/blob/d465196ab6d9ec6787ccd9e8643e69396928d2ba/pkg/addon/certpolicy/agent_addon.go#L37
> 
> ```go
> func NewGetValuesFunc(clusterClient  clusterclientset.Interface) addonfactory.GetValuesFunc {
> 	return func(cluster *clusterv1.ManagedCluster,
> 		addon *addonapiv1alpha1.ManagedClusterAddOn) (addonfactory.Values, error) {
> 		userValues := policyaddon.UserValues{
> 			GlobalValues: policyaddon.GlobalValues{
> 				managementClusterKubeVersion string
> 				...
> 		}
> 		if cluster is hosting mode {
> 			managementCluster,err:= clusterClient.ClusterV1().ManagedClusters().Get(managementCluster)
> 			userValues.GlobalValues.managementClusterKubeVersion = managementCluster.Status.Version.Kubernetes
> 		}
> 
> 		return addonfactory.JsonStructToValues(userValues)
> 	}
> }
> ```

@zhiweiyin318 sure, but do you not want to have this supported in the addon framework for all addons to inherit?

zhiweiyin318:
@mprahl  yes, we have to support this , but looks like lots of work need to consider and refactor.  for example, we need to consider update the existing  build-in values  https://github.com/open-cluster-management-io/addon-framework/blob/0325842035f67a1a1c120419bdfc4db9d7fe80b6/pkg/addonfactory/helm_agentaddon.go#L26 
not only the KubeVersion of management cluster, but also the AddonInstallNamespace may be different in management and managed clusters. and we also need to consider if we want to add new build-in values in the future.

How about to add a clusterClient to the `HelmAgentAddon` struct https://github.com/open-cluster-management-io/addon-framework/blob/0325842035f67a1a1c120419bdfc4db9d7fe80b6/pkg/addonfactory/helm_agentaddon.go#L42 and get the hosting cluster in ` (a *HelmAgentAddon) Manifests`

I'm concerned about the impact of the `Manifest` interface change on upgrades. 
https://github.com/open-cluster-management-io/addon-framework/pull/149/files#diff-937799135f99181b84091195a2d050345ae9cc5d51053825619a8672ff459f5fR35
I personally feel that the changes of `addonFactory` have less impact than the change of interface `Manifest`.

qiujian16:
My feeling is we should think more on the Manifest interface change. I am currently still wondering if we have another use case that we need to add general options variables on manifest interface in addition to feed management cluster kube version (And most of the addon do not need this version either). I 'd rather we have a fix with small impact on interface for now, and try to figure out whether we need and how to define such general options.

mprahl:
@zhiweiyin318 and @qiujian16 could you please take another look?

zhiweiyin318:
thanks @mprahl . looks good for me.
 @qiujian16 please take a look.

pkg/addonfactory/addonfactory.go
```diff
@@ -115,6 +116,13 @@ func (f *AgentAddonFactory) WithConfigGVRs(gvrs ...schema.GroupVersionResource)
 	return f
 }
 
+// WithHostingCluster defines the hosting cluster used in hosted mode. An AgentAddon may use this to provide
+// additional metadata.
+func (f *AgentAddonFactory) WithHostingCluster(cluster *clusterv1.ManagedCluster) *AgentAddonFactory {
+	f.hostingCluster = cluster
```
qiujian16, line 122: I think in sd case, there could be multiple hosting clusters? In the Manifest func, you could get the hosting cluster for the particular addon and then get that hosting cluster.

mprahl: @qiujian16 it seems that it is expected to have an `AgentAddonFactory` instance per addon based on the `addonName` parameter being required in `NewAgentAddonFactory`.

Since an addon can only be hosted on one hosting cluster, I think the approach in my PR is okay.

Ideally, I would have wanted to populate this automatically based on if the addon specified a hosting cluster, however, `AgentAddonFactory` doesn't have a Kubernetes client and I didn't want to introduce a breaking change of requiring it.

qiujian16: I think you could have one hub (where hub addon manager run), and multiple hosting clusters (where many hosted addon agents run on each hosting cluster and they all connect to the same hub)? If it is not the case, I am fine with this fix.

mprahl: @qiujian16 that's possible, but I'm saying there is a 1x1 mapping of `AgentAddonFactory` to addon, so  `AgentAddonFactory`  needs to only be aware of one hosting cluster.

qiujian16: @zhiweiyin318 do we have this requirement? do we have doc this?

mprahl: @qiujian16 I say it's a 1x1 mapping because of the function signature to create the `AgentAddonFactory` requires the addon name:
https://github.com/open-cluster-management-io/addon-framework/blob/259bd501bbfb2f22256073eb1c4e2fbf3bc754f6/pkg/addonfactory/addonfactory.go#L41

qiujian16: but addon name is per addon on hub I think. if you have multiple hosting clusters and one hub, you could have one addonfactory with a single addon name I think? Or maybe I miss something here.

mprahl: @qiujian16 oh, I see what you are saying. Yes, you are right because the ManagedClusterAddOn object is passed to the `Manifests` method and not to the addon factory.

With that in mind, I'm not quite sure how we could provide this change without changing the signature of a public API.

What do you suggest?

qiujian16: I think for now in the Manifests() func of helm agent addon factory, we could build a cluster client, get the hosting cluster and its version. Or even better, we could provide a `ManifestRenderingHook` on the factory, so the addon can read the hosting cluster in the hook func and render the manifest. It is not a perfect solution but I think it is better to defer the signature change until we have more use cases. 

I am thinking a general `ManifestOptions` as the input of Manifest() func, but I cannot find other use case in my to mind to use this option except hosting clusters (and even for hosting cluster, not all addon needs to know the version of hosting cluster). This is reason I feel we should not change signature for now.

Issue #134, opened 2022-11-09
`AddOnDeloyment` is missing a `p`


dhaiducek:
`AddOnDeloyment` --> `AddOnDeployment`

Affects:
- `GetAddOnDeloymentConfigValues`
- `NewAddOnDeloymentConfigGetter`

Not sure the best route for handling this, but a deprecation note and having the misspelled function point to the new function might work? It's a minor thing, but I think fixing it is important.

zhiweiyin318:
fixed.
/close

PR #152, opened 2023-02-21
fix typo AddOnDeloyment

zhiweiyin318:
fix the issue https://github.com/open-cluster-management-io/addon-framework/issues/134


qiujian16 (commented):
directly renaming will be breaking changes, you probably want to still keep the previous as a func aliase and mark them as deprecated.

zhiweiyin318:
> directly renaming will be breaking changes, you probably want to still keep the previous as a func aliase and mark them as deprecated.

@qiujian16  duplicated the typo function since the there is typo in the input and interface definition.  plz take a look again.

PR #160, opened 2023-03-16
fix gosec Implicit memory aliasing in for loop.

zhiweiyin318:


zhujian7:
missed one:
```
pkg/addonfactory/trimcrds.go:114:24: G601: Implicit memory aliasing in for loop. (gosec)
			removeDescriptionV1(&v)
```

PR #163, opened 2023-03-17
update cma config reference 

haoqing0110:
this PR is to update the config reference in cma status, it is part of the implementation of rollout strategy

- add manager generate default config references and install progression based on cma.spec.
- add framework update spec hash of the config references in cma status.


pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func (c *clusterManagementAddonConfigController) buildConfigInformers(
+	configInformerFactory dynamicinformer.DynamicSharedInformerFactory,
+	configGVRs map[schema.GroupVersionResource]bool,
+) []factory.Informer {
+	configInformers := []factory.Informer{}
+	for gvr := range configGVRs {
+		genericInformer := configInformerFactory.ForResource(gvr)
+		indexInformer := genericInformer.Informer()
```
qiujian16: is it necessary to put in a separated line?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,339 @@
...
+
+func (c *clusterManagementAddonConfigController) enqueueClusterManagementAddOnsByConfig(gvr schema.GroupVersionResource) enqueueFunc {
+	return func(obj interface{}) {
+		name, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("error to get accessor of object: %v", obj))
+			return
+		}
+
+		objs, err := c.clusterManagementAddonIndexer.ByIndex(byClusterManagementAddOnConfig, fmt.Sprintf("%s/%s/%s", gvr.Group, gvr.Resource, name))
```
qiujian16: does it also need the namespace?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,339 @@
...
+	if config.DesiredConfig.Namespace != "" {
+		return fmt.Sprintf("%s/%s/%s/%s", config.Group, config.Resource, config.DesiredConfig.Namespace, config.DesiredConfig.Name)
+	}
+
+	return fmt.Sprintf("%s/%s/%s", config.Group, config.Resource, config.DesiredConfig.Name)
+}
+
+func getInstallConfigIndex(config addonapiv1alpha1.InstallConfigReference) string {
+	if config.DesiredConfig.Namespace != "" {
+		return fmt.Sprintf("%s/%s/%s/%s", config.Group, config.Resource, config.DesiredConfig.Namespace, config.DesiredConfig.Name)
```
qiujian16: i think the index for empty namespace could "foo/bar//name". This is still unique for each one, right?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,339 @@
...
+		for j, configReference := range installProgression.ConfigReferences {
+			if configReference.DesiredConfig == nil || configReference.DesiredConfig.Name == "" {
+				continue
+			}
+
+			var err error
+			group := configReference.Group
+			resource := configReference.Resource
+			namespace := configReference.DesiredConfig.Namespace
+			name := configReference.DesiredConfig.Name
```
qiujian16: these lines seems unnecessary, we do not need to declare new variable.

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,339 @@
...
+
+func (c *clusterManagementAddonConfigController) getConfigSpecHash(group, resource, namespace, name string) (string, error) {
+	lister, ok := c.configListers[toListerKey(group, resource)]
+	if !ok {
+		return "", nil
+	}
+
+	var config *unstructured.Unstructured
+	var err error
+	if namespace == "" {
```
qiujian16: this if is not necessary

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	if err != nil {
+		return "", err
+	}
+
+	return GetSpecHash(config)
+}
+
+func getDefaultConfigIndex(config addonapiv1alpha1.DefaultConfigReference) string {
```
qiujian16: this seems duplicated with getInstallConfigIndex, can we reuse the same func?

pkg/manager/controllers/managementaddonstatus/controller.go
```diff
@@ -0,0 +1,229 @@
...
+	_, err = c.addonClient.AddonV1alpha1().ClusterManagementAddOns().Patch(
+		ctx, new.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
+	return err
+}
+
+func setDefaultConfigReference(supportedConfigs []addonv1alpha1.ConfigMeta,
+	existDefaultConfigReferences []addonv1alpha1.DefaultConfigReference) []addonv1alpha1.DefaultConfigReference {
+	newDefaultConfigReferences := []addonv1alpha1.DefaultConfigReference{}
+	for _, config := range supportedConfigs {
+		if config.DefaultConfig != nil {
```
qiujian16: I would do 

```
if config.DefaultConfig == nil {
    continue
}
```

to reduce nested

pkg/manager/controllers/managementaddonstatus/controller.go
```diff
@@ -0,0 +1,229 @@
...
+				configRef.DesiredConfig.SpecHash = existDefaultConfigReferences[i].DesiredConfig.SpecHash
+			}
+			newDefaultConfigReferences = append(newDefaultConfigReferences, configRef)
+		}
+	}
+	return newDefaultConfigReferences
+}
+
+func findDefaultConfigReference(newobj addonv1alpha1.DefaultConfigReference, oldobjs []addonv1alpha1.DefaultConfigReference) (int, bool) {
+	for i, oldconfig := range oldobjs {
```
qiujian16: I would return the object pointer instead of the index.

pkg/manager/controllers/managementaddonstatus/controller.go
```diff
@@ -0,0 +1,229 @@
...
+
+		// set config reference
+		installConfigReferences := []addonv1alpha1.InstallConfigReference{}
+		installConfigReferencesMap := map[addonv1alpha1.ConfigGroupResource]addonv1alpha1.ConfigReferent{}
+		for _, config := range supportedConfigs {
+			if config.DefaultConfig != nil {
+				installConfigReferencesMap[config.ConfigGroupResource] = *config.DefaultConfig
+			}
+		}
+		for _, config := range placementStrategy.Configs {
```
qiujian16, line 180: Add som doc here, it override the default configuration for each placement

pkg/manager/controllers/managementaddonstatus/controller.go
```diff
@@ -0,0 +1,229 @@
...
+		// if the config already exists in status, keep the existing spec hash
+		if i, exist := findInstallProgression(installProgression, existInstallProgressions); exist {
+			mergeInstallProgression(installProgression, existInstallProgressions[i])
+		}
+		newInstallProgressions = append(newInstallProgressions, installProgression)
+	}
+	return newInstallProgressions
+}
+
+func findInstallProgression(newobj addonv1alpha1.InstallProgression, oldobjs []addonv1alpha1.InstallProgression) (int, bool) {
```
qiujian16: return object pointer

pkg/addonmanager/controllers/managementaddonconfig/controller.go
haoqing0110: the name is from `cache.DeletionHandlingMetaNamespaceKeyFunc(obj)`, its value is actually `namespace/name` or name, may need a better var name.

pkg/addonmanager/controllers/managementaddonconfig/controller.go
haoqing0110: yes, still unique for each one, to check if namespace is empty or not is to make it easy to use `cache.DeletionHandlingMetaNamespaceKeyFunc(obj)` to get above `namespace/name` or `name` I think.

qiujian16: to may understanding DeletionHandlingMetaNamespaceKeyFunc can also handle /name as name, the returned namespace is just empty

pkg/addonmanager/manager.go
```diff
@@ -171,13 +172,20 @@ func (a *addonManager) Start(ctx context.Context) error {
 	)
 
 	var addonConfigController factory.Controller
+	var managementAddonConfigController factory.Controller
 	if len(a.addonConfigs) != 0 {
 		addonConfigController = addonconfig.NewAddonConfigController(
```
qiujian16, line 177: do we still need to keep this controller if addon manager is enabled?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,314 @@
...
+	configInformerFactory dynamicinformer.DynamicSharedInformerFactory,
+	configGVRs map[schema.GroupVersionResource]bool,
+) factory.Controller {
+	syncCtx := factory.NewSyncContext(controllerName)
+
+	c := &clusterManagementAddonConfigController{
+		addonClient:                   addonClient,
+		clusterManagementAddonLister:  clusterManagementAddonInformers.Lister(),
+		clusterManagementAddonIndexer: clusterManagementAddonInformers.Informer().GetIndexer(),
+		configListers:                 map[string]dynamiclister.Lister{},
```
qiujian16: can we use map[schema.GroupResource]dynamiclister.Lister?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
qiujian16: let's call it namespaceName

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,314 @@
...
+			utilruntime.HandleError(fmt.Errorf("error to get addons: %v", err))
+			return
+		}
+
+		for _, obj := range objs {
+			if obj == nil {
+				continue
+			}
+
+			cma := obj.(*addonapiv1alpha1.ClusterManagementAddOn)
```
qiujian16: you do not need this line for MetaNamespaceKeyFunc, or you want to log an error when type is not cma?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,314 @@
...
+			if configReference.DesiredConfig == nil || configReference.DesiredConfig.Name == "" {
+				// bad config reference, ignore
+				continue
+			}
+
+			configNames.Insert(getIndex(configReference.ConfigGroupResource, *configReference.DesiredConfig))
+		}
+	}
+
+	if configNames.Len() == 0 {
```
qiujian16: seems not needed, empty will return an empty list, right?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -0,0 +1,314 @@
...
+
+func (c *clusterManagementAddonConfigController) updateConfigSpecHash(cma *addonapiv1alpha1.ClusterManagementAddOn) error {
+
+	for i, defaultConfigReference := range cma.Status.DefaultConfigReferences {
+		if defaultConfigReference.DesiredConfig == nil || defaultConfigReference.DesiredConfig.Name == "" {
+			continue
+		}
+
+		var err error
+		cma.Status.DefaultConfigReferences[i].DesiredConfig.SpecHash, err = c.getConfigSpecHash(defaultConfigReference.Group,
```
qiujian16: I think we should directly set specHash, but check whether there is err, Otherwise this specHash will be set an empty string. I know we return error after that, but we should avoid directly setting here.

pkg/addonmanager/manager.go
haoqing0110: yes, the addonConfigController is responsible to update spec hash for mca.spec.configs when addon manager is enabled.

PR #164, opened 2023-03-29
update mca config reference 

haoqing0110:
this PR is to update the config reference in mca status, it is part of the implementation of rollout strategy

- add manager generate config references for each mca, based on cma.spec. mca.spec.
- add manager sync spec hash from cma status to mca status config reference. 
- add framework update spec hash of the mca.spec.configs in mca status config reference.

### DIFF
--- a/pkg/addonmanager/controllers/addonconfig/controller.go
+++ b/pkg/addonmanager/controllers/addonconfig/controller.go
@@ -8,7 +8,6 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -30,9 +29,8 @@ import (
 )
 
 const (
-	controllerName               = "addon-config-controller"
-	byAddOnConfig                = "by-addon-config"
-	UnsupportedConfigurationType = "UnsupportedConfiguration"
+	controllerName = "addon-config-controller"
+	byAddOnConfig  = "by-addon-config"
 )
 
 type enqueueFunc func(obj interface{})
@@ -163,13 +161,6 @@ func (c *addonConfigController) sync(ctx context.Context, syncCtx factory.SyncCo
 		return err
 	}
 
-	// if there is unsupported configs defined in spec, skip updating spec hash.
-	configCond := meta.FindStatusCondition(addon.Status.Conditions, UnsupportedConfigurationType)
-	if configCond == nil || configCond.Status == metav1.ConditionTrue {
-		klog.Warningf("failed to update spec hash as unsupported configuration defined in %s", addon.Name)
-		return nil
-	}
-
 	addonCopy := addon.DeepCopy()
 
 	if err := c.updateConfigSpecHashAndGenerations(addonCopy); err != nil {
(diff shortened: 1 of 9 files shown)


pkg/addonmanager/controllers/managementaddonconfig/controller.go
```diff
@@ -266,7 +266,13 @@ func (c *clusterManagementAddonConfigController) getConfigSpecHash(gr addonapiv1
 		return "", nil
 	}
 
-	config, err := lister.Namespace(cr.Namespace).Get(cr.Name)
+	var config *unstructured.Unstructured
+	var err error
+	if cr.Namespace == "" {
```
qiujian16, line 271: you do not need this part to check namespace.

pkg/manager/controllers/addonconfiguration/addon_configuration_reconciler.go
```diff
@@ -59,7 +59,13 @@ func (d *managedClusterAddonConfigurationReconciler) buildConfigurationGraph(cma
 			continue
 		}
 
-		graph.addPlacementNode(strategy.Configs, clusters)
+		installConfigReferences := []addonv1alpha1.InstallConfigReference{}
+		for _, installProgression := range cma.Status.InstallProgressions {
+			if installProgression.PlacementRef == strategy.PlacementRef {
```
qiujian16: will it be racy? When spec is updated but status is not updated yet?

pkg/manager/controllers/addonconfiguration/addon_configuration_reconciler.go
```diff
@@ -131,8 +137,9 @@ func (d *managedClusterAddonConfigurationReconciler) mergeAddonConfig(
 			}
 
 			match = true
-			if mergedConfigs[i].ConfigReferent != config.ConfigReferent {
+			if !equality.Semantic.DeepEqual(mergedConfigs[i].DesiredConfig, config.DesiredConfig) {
```
qiujian16: what happens if configReferent is the same but desiredConfig is different? should we not clean the generation?

pkg/addonmanager/controllers/addonconfig/controller.go
```diff
@@ -165,36 +163,35 @@ func (c *addonConfigController) sync(ctx context.Context, syncCtx factory.SyncCo
 		return err
 	}
 
+	// if there is unsupported configs defined in spec, skip updating spec hash.
+	configCond := meta.FindStatusCondition(addon.Status.Conditions, UnsupportedConfigurationType)
```
qiujian16: I think this is just a hint for the user. You could have one config supported and another unsupported, we should still update the specHash of the supported config in this case, right?

pkg/addonmanager/controllers/managementaddonconfig/controller.go
haoqing0110: add this check back since for cluster scope resource, `config, err = lister.Namespace(cr.Namespace).Get(cr.Name)` can't get the config.

pkg/manager/controllers/addonconfiguration/addon_configuration_reconciler.go
haoqing0110: yes, this part need modification, should keep the generation when configReferent (name/namespace) is the same and only config spec hash changes.

pkg/addonmanager/controllers/addonconfig/controller.go
haoqing0110: This function is for the framework to update the spec hash for mca.spec.configs. I'm thinking the case when at time1 user defines supported config A, and the status has spec hash for config A. Then at time2, user add unsupported configB in mca.spec.configs, and modify config A, should we continue update the spec hash for config A, and let it synced to manifestwork? or do not update anything.

qiujian16: I think we should honor config A in this case.

pkg/manager/controllers/addonconfiguration/addon_configuration_reconciler.go
haoqing0110: yes, should only sync data from cma.status.

haoqing0110: when spec is updated and status is not, might sync old spec hash to mca.staus. Maybe a better way is to just sync cma.status.InstallProgressions to mca.

qiujian16: yes I think so

haoqing0110: fixed in latest commit.

pkg/manager/controllers/addonconfiguration/addon_configuration_reconciler.go
haoqing0110: fixed in latest commit.

pkg/addonmanager/controllers/addonconfig/controller.go
haoqing0110: fixed in latest commit.

PR #165, opened 2023-04-01
Handle update conflict in agentdeploy controller

tpantelis:
We see these periodic error messages logged:

`base_controller.go:159] "addon-deploy-controller" controller failed to sync "prod-on-prem/submariner", err: Operation cannot be fulfilled on managedclusteraddons.addon.open-cluster-management.io "submariner": the object has been modified; please apply your changes to the latest version and try again`

Assuming no other controller is making frequent updates to the `ManagedClusterAddon`, the most likely reason for the update conflicts is due to caching. The controller obtains the `ManagedClusterAddon` from the informer's cache so it may not get the most up-to-date version of the resource which can cause the transient conflict errors when updating the resource. This can happen if the resource is immediately re-queued after an update and the updated version hasn't reached the cache yet when the controller's sync function is invoked again.

The log messages are benign but they are disconcerting and can fill up the log with noise. To avoid the logging, on conflict error, simply re-queue the resource instead of returning the error.

### DIFF
--- a/pkg/basecontroller/factory/base_controller.go
+++ b/pkg/basecontroller/factory/base_controller.go
@@ -6,6 +6,8 @@ import (
 	"sync"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
@@ -155,7 +157,11 @@ func (c *baseController) processNextWorkItem(queueCtx context.Context) {
 	}
 
 	if err := c.sync(queueCtx, syncCtx, queueKey); err != nil {
-		if klog.V(4).Enabled() || key != "key" {
+		err = tryUnwrapAggregate(err)
+
+		if apierrors.IsConflict(err) {
+			klog.V(4).Infof("%q controller encountered a conflict - re-queueing: %v", c.name, err)
+		} else if klog.V(4).Enabled() || key != "key" {
 			utilruntime.HandleError(fmt.Errorf("%q controller failed to sync %q, err: %w", c.name, key, err))
 		} else {
 			utilruntime.HandleError(fmt.Errorf("%s reconciliation failed: %w", c.name, err))
@@ -165,4 +171,13 @@ func (c *baseController) processNextWorkItem(queueCtx context.Context) {
 	}
 
 	c.syncContext.Queue().Forget(key)
+}
+
+func tryUnwrapAggregate(err error) error {
+	a, ok := err.(utilerrors.Aggregate)
+	if !ok || len(a.Errors()) != 1 {
+		return err
+	}
+
+	return tryUnwrapAggregate(a.Errors()[0])
 }


pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -219,8 +219,14 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 	}
 
 	if err = c.updateAddon(ctx, addon, oldAddon); err != nil {
+		if errors.IsConflict(err) {
+			syncCtx.Queue().AddRateLimited(key)
+			return nil
```
zhiweiyin318: if return nil,  https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/basecontroller/factory/base_controller.go#L167  this place will forget the key.
I think it may be better to check if conflict  here https://github.com/open-cluster-management-io/addon-framework/blob/549a39952992436903c51bf3849d8530271a1654/pkg/basecontroller/factory/base_controller.go#L158

tpantelis: yeah good point. It looks like `base_controller` et al was copied from  [openshift/library-go](https://github.com/openshift/library-go/tree/master/pkg/controller/factory).  I thought about doing the check in `base_controller` but didn't want it getting lost in case the project was ever changed to use `openshift/library-go`. I think the best thing to do is to introduce [SyntheticRequeueError](https://github.com/openshift/library-go/blob/6ed98e0ed0b908793ef4958197eb1a1c0577aa23/pkg/controller/factory/base_controller.go#L265).

qiujian16: we actually switch from openshift/library-go. The reason is we would like to avoid the deps on library-go which has many conflicts with controller-runtime before, and many addons use controller-runtime. So we copy the controller library code only from library-go but remove unnecessary deps. 

I think we could tread conflict as a special error with different log level if needed.

tpantelis: OK, if we handle conflict error in `base_controller` then every client controller will get the behavior as well. I haven't looked at all the other controllers using `base_controller` but if you think there won't be any issues then I'll change `base_controller`.

qiujian16: yes I think we could log error when log level is 4 and it is conflict error, I think we still need to surface the conflict error because it may imply a bug in the code.

tpantelis:
@zhiweiyin318 @qiujian16 An [integration test](https://github.com/open-cluster-management-io/addon-framework/actions/runs/4625472917/jobs/8203569767?pr=165#step:4:210.) failed.  I don't see how my changes could've caused that, plus they all succeed locally for me. Has there been past issues with flaky test failures? Unfortunately I don't have the ability to re-run the job (not that I can see).

zhiweiyin318:
@tpantelis  We are trying to merge controllers to reduce conflicts. If the result is good, we can avoid modifying base_controller.

zhiweiyin318:
@tpantelis  I refactored agentDeploy controller [PR-172](https://github.com/open-cluster-management-io/addon-framework/pull/172) and fix issue in work applier [PR-233](https://github.com/open-cluster-management-io/api/pull/233) . the most of conflict has been avoided.  please update the latest addon-framework lib to have a try.

PR #166, opened 2023-04-04
Start addon status update in addon's own manager

qiujian16:
if addon lifecycle annotation is not set or the value is self, addon's own status update controller should process


pkg/addonmanager/manager.go
```diff
@@ -171,8 +177,15 @@ func (a *addonManager) Start(ctx context.Context) error {
 		a.addonAgents,
 	)
 
-	var addonConfigController factory.Controller
-	var managementAddonConfigController factory.Controller
+	// This is a duplicate controller in general addon-manager. This should be removed when we
+	// alway enable the addon-manager
+	addonStatusController := addonstatus.NewAddonStatusController(
```
haoqing0110: Does this status controller also need to be enabled by annotation?

qiujian16: I think it is unnecessary, we can start this with noop. enable by annotation is weird, because user actually can add/remove annotation, and we have to watch the resource and dynamically start/stop this controller

pkg/utils/helpers.go
```diff
@@ -282,3 +282,36 @@ func PatchAddonCondition(ctx context.Context, addonClient addonv1alpha1client.In
...
+	value, ok := cma.Annotations[addonapiv1alpha1.AddonLifecycleAnnotationKey]
+	if !ok {
+		return false
+	}
+
+	return value == addonapiv1alpha1.AddonLifecycleAddonManagerAnnotationValue
+}
+
+func ManagedBySelf(cma *addonapiv1alpha1.ClusterManagementAddOn) bool {
+	return !ManagedByAddonManager(cma)
```
haoqing0110: Do we need to limit the user can only define annotation value to `"self"` and `addon-manager`, or anything other than `addon-manager` will be treated as self?

pkg/manager/controllers/addonowner/controller.go
```diff
@@ -0,0 +1,94 @@
...
+		addonFilterFunc:              addonFilterFunc,
+	}
+
+	return factory.New().WithInformersQueueKeysFunc(
+		func(obj runtime.Object) []string {
+			key, _ := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			return []string{key}
+		},
+		addonInformers.Informer(), clusterManagementAddonInformers.Informer()).
+		WithSync(c.sync).ToController("addon-status-controller")
```
haoqing0110: should this rename to addon-owner-controller.

PR #170, opened 2023-04-13
update mca progressing condition

haoqing0110:
Update mca progressing condition


pkg/manager/controllers/addonprogressing/controller.go
```diff
@@ -0,0 +1,281 @@
...
+type addonProgressingController struct {
+	addonClient               addonv1alpha1client.Interface
+	managedClusterAddonLister addonlisterv1alpha1.ManagedClusterAddOnLister
+	workLister                worklister.ManifestWorkLister
+}
+
+func NewAddonProgressingController(
+	addonClient addonv1alpha1client.Interface,
+	addonInformers addoninformerv1alpha1.ManagedClusterAddOnInformer,
+	workInformers workinformers.ManifestWorkInformer,
```
qiujian16, line 54: we also need to add the filter that some addon will not be handled by this manager

pkg/manager/controllers/addonprogressing/controller.go
```diff
@@ -0,0 +1,281 @@
...
+	return c.updateAddonProgressingAndLastApplied(ctx, addon.DeepCopy(), addon)
+}
+
+func (c *addonProgressingController) updateAddonProgressingAndLastApplied(ctx context.Context, newaddon, oldaddon *addonapiv1alpha1.ManagedClusterAddOn) error {
+	// check config references
+	if supported, config := isConfigurationSupported(newaddon); !supported {
+		meta.SetStatusCondition(&newaddon.Status.Conditions, metav1.Condition{
+			Type:    addonapiv1alpha1.ManagedClusterAddOnConditionProgressing,
+			Status:  metav1.ConditionFalse,
+			Reason:  constants.ProgressingConfigurationUnsupported,
```
qiujian16: put this into api repo

pkg/manager/controllers/addonprogressing/controller.go
```diff
@@ -0,0 +1,281 @@
...
+			Message: fmt.Sprintf("Configuration with gvr %s/%s is not supported for this addon", config.Group, config.Resource),
+		})
+		return c.patchAddOnProgressingAndLastApplied(ctx, newaddon, oldaddon)
+	}
+
+	// get addon works
+	requirement, _ := labels.NewRequirement(addonapiv1alpha1.AddonLabelKey, selection.Equals, []string{newaddon.Name})
+	selector := labels.NewSelector().Add(*requirement)
+	addonWorks, err := c.workLister.ManifestWorks(newaddon.Namespace).List(selector)
```
qiujian16: add a todo here since we does not consider hosted manifestwork yet

pkg/manager/controllers/addonprogressing/controller.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil
+	}
+
+	addon, err := c.managedClusterAddonLister.ManagedClusterAddOns(namespace).Get(addonName)
+	switch {
+	case errors.IsNotFound(err):
+		return nil
+	case err != nil:
+		return err
+	}
```
qiujian16, line 96: It is better to wait until ManifestApplied condition is put on the addon

pkg/manager/controllers/addonprogressing/controller.go
```diff
@@ -0,0 +1,281 @@
...
+	// set upgrade flag
+	isUpgrade := false
+	for _, configReference := range newaddon.Status.ConfigReferences {
+		if configReference.LastAppliedConfig != nil && configReference.LastAppliedConfig.SpecHash != "" {
+			isUpgrade = true
+			break
+		}
+	}
+
+	if len(addonWorks) == 0 {
```
qiujian16, line 157: should the deploy controller wait for progressing state? I think it should

pkg/manager/controllers/managementaddonprogressing/controller.go
```diff
@@ -0,0 +1,238 @@
...
+			return []string{accessor.GetName()}
+		},
+		addonInformers.Informer(), clusterManagementAddonInformers.Informer()).
+		WithInformersQueueKeysFunc(index.ClusterManagementAddonByPlacementDecisionQueueKey(clusterManagementAddonInformers), placementDecisionInformer.Informer()).
+		WithInformersQueueKeysFunc(index.ClusterManagementAddonByPlacementQueueKey(clusterManagementAddonInformers), placementInformer.Informer()).
+		WithSync(c.sync).ToController("management-addon-status-controller")
+
+}
+
+func (c *managementAddonProgressingController) sync(ctx context.Context, syncCtx factory.SyncContext, addonName string) error {
```
qiujian16: how do you think about putting this as a reconciler in addonConfiguration reconciler? It seems to me that since we have the graph built, getting to know the number of updating and set condition is easier?

pkg/manager/controllers/addonprogressing/controller.go
```diff
@@ -0,0 +1,281 @@
...
+
+// set addon progressing condition
+func setAddOnProgressing(isUpgrade, isSucceed bool, addon *addonapiv1alpha1.ManagedClusterAddOn) {
+	condition := metav1.Condition{
+		Type: addonapiv1alpha1.ManagedClusterAddOnConditionProgressing,
+	}
+	if !isSucceed {
+		condition.Status = metav1.ConditionTrue
+		if isUpgrade {
+			condition.Reason = constants.ProgressingUpgrading
```
qiujian16: reasons also should be put in api repo

pkg/manager/controllers/addonprogressing/controller.go
haoqing0110: add the check after checking supported configs and before setting other progressing reasons.

pkg/manager/controllers/addonprogressing/controller.go
haoqing0110: added

pkg/manager/controllers/addonprogressing/controller.go
haoqing0110: added in deploy controller

pkg/manager/controllers/addonprogressing/controller.go
haoqing0110: added filter.

pkg/manager/controllers/addonprogressing/controller.go
haoqing0110: Needs further discussion here, since progressing state is only added when cma is managed by addon-manager, while deploy controller is always running by the addon itself.

pkg/manager/controllers/managementaddonprogressing/controller.go
haoqing0110: remove managementaddonprogressing from this PR. will create another PR to update cma progressing.

PR #172, opened 2023-04-13
refactor healthcheck controller into agentdeploy controller to reduce conflicts

zhiweiyin318:


pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -169,6 +172,11 @@ func (c *addonDeployController) sync(ctx context.Context, syncCtx factory.SyncCo
 		return err
 	}
 
+	// to deploy agents after RegistrationApplied is true
+	if !meta.IsStatusConditionTrue(addon.Status.Conditions, addonapiv1alpha1.ManagedClusterAddOnRegistrationApplied) {
```
qiujian16: we should only check the existence of the condition

pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
```diff
@@ -0,0 +1,177 @@
...
+		meta.SetStatusCondition(&addon.Status.Conditions, metav1.Condition{
+			Type:    addonapiv1alpha1.ManagedClusterAddOnConditionAvailable,
+			Status:  metav1.ConditionTrue,
+			Reason:  addonapiv1alpha1.AddonAvailableReasonWorkApply,
+			Message: "Addon manifestWork is applied",
+		})
+		return nil
+	}
+
+	if !meta.IsStatusConditionTrue(addon.Status.Conditions, addonapiv1alpha1.ManagedClusterAddOnManifestApplied) {
```
qiujian16: I think maybe we should only adding conditions of available when there is manifestapplied condition

qiujian16:
we should also skip AlreadyExists error during create.

zhiweiyin318:
> we should also skip AlreadyExists error during create.

@qiujian16  added

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -232,14 +244,54 @@ func (c *addonDeployController) updateAddon(ctx context.Context, new, old *addon
...
+		ctx, new.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
+	return err
 }
 
 func (c *addonDeployController) applyWork(ctx context.Context, appliedType string,
 	work *workapiv1.ManifestWork, addon *addonapiv1alpha1.ManagedClusterAddOn) (*workapiv1.ManifestWork, error) {
 
 	work, err := c.workApplier.Apply(ctx, work)
 	if err != nil {
+		if errors.IsAlreadyExists(err) {
```
qiujian16: will this return "nil, nil"? should we change this in applier instead of here?

zhiweiyin318: removed the check ,added into applier

PR #177, opened 2023-04-24
add addon filter for addonmanger

zhiweiyin318:


pkg/manager/controllers/addonowner/controller.go
```diff
@@ -80,9 +86,10 @@ func (c *addonOwnerController) sync(ctx context.Context, syncCtx factory.SyncCon
 	}
 
 	if !c.addonFilterFunc(clusterManagementAddon) {
+		klog.Infof("\n ### cma not filtered: %v+", clusterManagementAddon)
 		return nil
 	}
-
+	klog.Infof("\n ### cma filtered: %v+", clusterManagementAddon)
```
qiujian16: this line is unnecessary

zhiweiyin318: removed

PR #178, opened 2023-04-26
Make manifestWork UpdateStrategy configurable through AgentAddonOptions

### LINKED ISSUE #173: RFE: UpdateStrategy in AgentAddonOptions
Today it is not possible to specify in [AgentAddonOptions](https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/agent/inteface.go#L42-L73) the [UpdateStrategy](https://github.com/open-cluster-management-io/api/blob/v0.10.1/work/v1/types.go#L97). 
It would be useful to specify that "create only" or  "server side apply" is to be used for some manifests containing configuration data so that regular changes made by users do not get overwritten by an update of the resource pushed by the addon as described here:
https://open-cluster-management.io/concepts/manifestwork/#resource-race-and-adoption

If you find this request meaningful I am happy to try to contribute the implementation.

fgiloux:
It makes possible to specify that "create only" or "server side apply" is to be used for some manifests containing configuration data so that regular changes made by users do not get overwritten by an update of the resource pushed by the addon.

fixes: #173

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/utils.go
+++ b/pkg/addonmanager/controllers/agentdeploy/utils.go
@@ -474,26 +474,29 @@ func newAddonWorkObjectMeta(namePrefix, addonName, addonNamespace, workNamespace
 }
 
 func getManifestConfigOption(agentAddon agent.AgentAddon) []workapiv1.ManifestConfigOption {
-	if agentAddon.GetAgentAddonOptions().HealthProber == nil {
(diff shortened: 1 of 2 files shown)


fgiloux:
cc @zhiweiyin318 @qiujian16

qiujian16 (commented):
/approve

looks good to me
/assign @zhiweiyin318

PR #184, opened 2023-05-23
only enable install strategy and update install progression when addon is managed by addon-manager

haoqing0110:
This PR tried to fix: 
1 When add-on cma is managed by self, cma defined install strategy should not work, since it has conflicted with hard code WithInstallStrategy().
2 When add-on cma is managed by self, cma should not list installprogression in status.


qiujian16 (commented):
/approve
/lgtm

PR #190, opened 2023-07-03
:sparkles: support addon health probe for checking deployment availability

zhujian7:
Add a new type to check addon health `DeploymentAvailability`, the addon status will be ok when all deployments deployed on the agent cluster are available.


qiujian16:
/approve
@zhiweiyin318 please review

zhiweiyin318:
Do we have plan to support more general healthcheck, like StatefulSets, DaemonSet?

Issue #198, opened 2023-07-28
The agentdeploy controller cannot update the agent manifestWorks in time when the cluster is changed


zhiweiyin318:
there are some cases the agent manifestworks have dependancies on the managed cluster, like the annotation or clusterClaim. 
but the agentdeploy controller only watches the addon and manifestwork, does not watch cluster, so cannot update manifestWorks in time when the cluster is changed.

zhiweiyin318:
cc @zhujian7

zhujian7:
It is not a good idea to watch the managed cluster for all addons, since it is easy to change, which may cause some performance issues. So we are going to provide a filter function that developers can use to filter which information the addon is interested in: https://github.com/open-cluster-management-io/addon-framework/pull/199

PR #199, opened 2023-07-28
make the agent deploy controller watch the managed cluster

zhujian7:
Provide an option for developer to make the agent deploy controller watches the managed cluster, so if there are any changes on the managed cluster, the agent deploy controller can reconcile quickly

for example: if a helm addon wants managed cluster annotation changing trigger the agent redeploy:
```
.WithAgentDeployTriggerClusterFilter(func(old, new *clusterv1.ManagedCluster) bool {
			return !equality.Semantic.DeepEqual(old.Annotations, new.Annotations)
		})
```

issue reference: https://github.com/open-cluster-management-io/addon-framework/issues/198

### DIFF
--- a/pkg/agent/inteface.go
+++ b/pkg/agent/inteface.go
@@ -76,6 +76,19 @@ type AgentAddonOptions struct {
 	// SupportedConfigGVRs is a list of addon supported configuration GroupVersionResource
 	// each configuration GroupVersionResource should be unique
 	SupportedConfigGVRs []schema.GroupVersionResource
+
+	// AgentDeployTriggerClusterFilter defines the filter func to trigger the agent deploy/redploy when cluster info is
+	// changed. Addons that need information from the ManagedCluster resource when deploying the agent should use this
+	// field to set what information they need, otherwise the expected/up-to-date agent may be deployed delayed since
+	// the default filter func returns false when the ManagedCluster resource is updated.
+	//
+	// For example, the agentAddon needs information from the ManagedCluster annotation, it can set the filter function
+	// like:
+	//
+	//	AgentDeployTriggerClusterFilter: func(old, new *clusterv1.ManagedCluster) bool {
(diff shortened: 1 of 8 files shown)


pkg/agent/inteface.go
```diff
@@ -76,6 +76,19 @@ type AgentAddonOptions struct {
...
+	// field to set what information they need, otherwise the expected/up-to-date agent may be deployed delayed since
+	// the default filter func returns false when the ManagedCluster resource is updated.
+	//
+	// For example, the agentAddon needs information from the ManagedCluster annotation, it can set the filter function
+	// like:
+	//
+	//	AgentDeployClusterInformerFilter: func(old, new *clusterv1.ManagedCluster) bool {
+	//	 return !equality.Semantic.DeepEqual(old.Annotations, new.Annotations)
+	//	}
+	AgentDeployClusterInformerFilter func(old, new *clusterv1.ManagedCluster) bool
```
qiujian16:
```suggestion
	AgentDeployTriggerClusterFilter func(old, new *clusterv1.ManagedCluster) bool
```
something like this?

pkg/agent/inteface.go
```diff
@@ -76,6 +76,19 @@ type AgentAddonOptions struct {
 	// SupportedConfigGVRs is a list of addon supported configuration GroupVersionResource
 	// each configuration GroupVersionResource should be unique
 	SupportedConfigGVRs []schema.GroupVersionResource
+
+	// AgentDeployClusterInformerFilter defines the filter function for the cluster informer when deploying the addon
```
qiujian16: the filter func to trigger the agent deploy/redploy when cluster info is changed.

zhujian7: fixed

pkg/agent/inteface.go
zhujian7: ok, fixed

PR #203, opened 2023-08-24
:warning: Prefer namespace defined in the registration option

zhujian7:


pkg/agent/inteface.go
```diff
@@ -108,8 +108,11 @@ type RegistrationOption struct {
...
 	CSRConfigurations func(cluster *clusterv1.ManagedCluster) []addonapiv1alpha1.RegistrationConfig
 
-	// Namespace is the namespace where registraiton credential will be put on the managed cluster. It
-	// will be overridden by installNamespace on ManagedClusterAddon spec if set
+	// Namespace is the namespace where registration credential will be put on the managed cluster. If it is not
+	// empty it will override the installNamespace on ManagedClusterAddon spec.
+	// Note: Set this very carefully. If this is set, the addon agent must be deployed in the same namespace, which
+	// means when implementing Manifests function in AgentAddon interface, the namespace of the addon agent must be
+	// set to this value as well.
 	Namespace string
```
zhujian7, line 116: I am considering if it is better to add a new filed `AgentNamespace` and deprecate the `Namespace` field
```
AgentNamespace  func(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn) (string, error)
```
 since the new field is a func, we can get the value from addontemplate when it is a template type addon.

@qiujian16 WDYT?

PR #205, opened 2023-09-05
add a helper func for setting agent install namespace from addon deployment config

zhujian7:
Add a helper func for setting the `registrationOption.AgentInstallNamespace`, This helper func will try to get the namespace from the addon deployment config. If the addon does not support addon deployment config or there is no matched addon deployment config, it will return an empty string.


pkg/utils/namespace.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		if config == nil {
+			return ""
+		}
+
+		return config.Spec.AgentInstallNamespace
+	}
+}
+
+// GetDesiredAddOnDeployment returns the desired template of the addon
```
qiujian16:
```suggestion
// GetDesiredAddOnDeployment returns the desired addonDeploymentConfig of the addon
```

pkg/utils/namespace.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/klog/v2"
+	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+)
+
+func AgentInstallNamespaceFromDeploymentConfig(
+	adcgetter func(ctx context.Context, namespace, name string) (*addonapiv1alpha1.AddOnDeploymentConfig, error),
```
qiujian16: think this one is supposed to be used in a controller to update status on addon?

How about we have a func to sync namespace in config to addon.status?

cmd/example/helloworld/main.go
```diff
@@ -93,6 +93,11 @@ func runController(ctx context.Context, kubeConfig *rest.Config) error {
 		utilrand.String(5),
 	)
 
+	// Set agent install namespace from addon deployment config if it exists
+	registrationOption.AgentInstallNamespace = utils.AgentInstallNamespaceFromDeploymentConfig(
```
qiujian16: I do not think we should only set during start

qiujian16 (commented):
/approve
/lgtm

PR #214, opened 2023-10-19
:sparkles: Support setting manifest work config for agents

zhujian7:
With this, the addonTemplate can set the `ManifestConfigs` for the generated manifestworks


pkg/agent/inteface.go
```diff
@@ -88,7 +88,14 @@ type AgentAddonOptions struct {
 	//	AgentDeployTriggerClusterFilter: func(old, new *clusterv1.ManagedCluster) bool {
 	//	 return !equality.Semantic.DeepEqual(old.Annotations, new.Annotations)
 	//	}
+	// +optional
 	AgentDeployTriggerClusterFilter func(old, new *clusterv1.ManagedCluster) bool
+
+	// ManifestConfigs represents the configurations of manifests defined in workload field.
+	// It will override the update strategy setted by the "Updaters" field abd merge the feedback rules setted by the
```
qiujian16:
```suggestion
	// It will override the update strategy set by the "Updaters" field and merge the feedback rules set by the
```

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -517,9 +518,45 @@ func getManifestConfigOption(agentAddon agent.AgentAddon,
...
+		if mc.UpdateStrategy != nil {
+			manifestConfigs[index].UpdateStrategy = mc.UpdateStrategy
+		}
+	}
 	return manifestConfigs
 }
 
+func containsResourceIdentifier(mcs []workapiv1.ManifestConfigOption, ri workapiv1.ResourceIdentifier) int {
+	for index, mc := range mcs {
+		if reflect.DeepEqual(mc.ResourceIdentifier, ri) {
```
qiujian16: you could directly use == for ResourceIdentifier I think. It is plain struct

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -517,9 +518,45 @@ func getManifestConfigOption(agentAddon agent.AgentAddon,
...
+			return index
+		}
+	}
+	return -1
+}
+
+func containsFeedbackRule(rules []workapiv1.FeedbackRule, rule workapiv1.FeedbackRule) bool {
+	for _, r := range rules {
+		if reflect.DeepEqual(r, rule) {
```
qiujian16: actually we may only compare the name of the rule. It won't work if you have two rules with the same name.

zhujian7: you mean the jsonpath.name? so if they have the same name but a different path, should we override the old one with the agentOption.ManifestConfigs?

qiujian16: Yes, we have to decide override or not, but comparing name only should be good enough.

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -518,18 +517,17 @@ func getManifestConfigOption(agentAddon agent.AgentAddon,
 		}
 	}
 
-	emc := agentAddon.GetAgentAddonOptions().ManifestConfigs
-	for _, mc := range emc {
+	aaomc := agentAddon.GetAgentAddonOptions().ManifestConfigs
```
qiujian16: what does aaomc stands for :). You might want a more meaningful name, event it is longer.

zhujian7: agentAddonOptionsManifestConfigs, but I changed it to userConfiguredManifestConfigs

Issue #217, opened 2023-10-29
Add support to configure helm-agent-addon namespace.


Imtiaz246:
**Problem Statement:**
- We are creating a `helm-agent-addon` using the [installStrategy feature](https://github.com/open-cluster-management-io/api/blob/bab1208e6889582db98e95e8210680c5ef843a5a/addon/v1alpha1/types_clustermanagementaddon.go#L61) defined in the `ClusterManagementAddon` API. Currently we can't configure the namespace dynamically as per the [ocm:issues:298](https://github.com/open-cluster-management-io/ocm/issues/298)  describes. In our use case, we need the ability to configure the namespace for the helm-agent-addon dynamically. 
- Additionally, we want to leverage helm chart built-in values such as `{{ Release.Namespace }}` and `{{ Release.Name }}` to set the values of the helm-agent-addon.

**Proposed Solution:**
- We propose adding a feature/enhancement that allows for the dynamic configuration of the `helm-agent-addon` namespace. This would enable users to specify the namespace at build time through some configuration parameters.

**Use Case Example:**
- For example, in our use case, we want to set the `helm-agent-addon` namespace using the helm chart built-in value `{{ Release.Namespace }}` which would ensure consistency with the parent Helm release.

zhujian7:
@Imtiaz246 thanks for reporting this

> we can't configure the namespace dynamically as per the https://github.com/open-cluster-management-io/ocm/issues/298 describes.

PR https://github.com/open-cluster-management-io/addon-framework/pull/218 is trying to fix this.


> additionally, we want to leverage helm chart built-in values such as {{ Release.Namespace }} and {{ Release.Name }}

there are [3 helm built-in values](https://open-cluster-management.io/developer-guides/addon/#values-definition) are supported, fortunately, `{{ Release.Namespace }}` and `{{ Release.Name }}` are included.

zhujian7:
@Imtiaz246 I submitted [a PR to support configuring the namespace for the helm addons](https://github.com/open-cluster-management-io/addon-framework/pull/218), if you want you can try it with the latest addon-framework, example [here](https://github.com/open-cluster-management-io/addon-framework/blob/main/cmd/example/helloworld_helm/main.go#L130-L135)

Imtiaz246:
Great @zhujian7. Thank you for your quick response. Things are working as expected now.

tamalsaha:
@zhujian7 , I think there is still one thing left to fix

[...]

PR #218, opened 2023-10-30
✨  Support configuring install namespace for helm addon

zhujian7:
issue reference: https://github.com/open-cluster-management-io/addon-framework/issues/217


qiujian16:
leave to @zhiweiyin318 for final review

PR #226, opened 2023-12-07
Continue updating deployment manifests when uninstalling

mprahl:
This allows an addon to update its deployment manifests while the predelete hook is running. This is useful when telling the addon to enter uninstall mode.

Additionally, if there is a bug in the uninstall and it keeps the uninstall from succeeding, we need to allow container image updates to move forward in the event the update fixes the problem.

Relates:
https://issues.redhat.com/browse/ACM-8947

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
+++ b/pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
@@ -80,34 +80,33 @@ func (s *hostedSyncer) sync(ctx context.Context,
 		Message: fmt.Sprintf("hosting cluster %s is a managed cluster of the hub", hostingClusterName),
 	})
 
-	if !hostingCluster.DeletionTimestamp.IsZero() {
-		if err = s.cleanupDeployWork(ctx, addon); err != nil {
-			return addon, err
-		}
-		addonRemoveFinalizer(addon, addonapiv1alpha1.AddonHostingManifestFinalizer)
-		return addon, nil
-	}
-
-	if !addon.DeletionTimestamp.IsZero() {
-		// clean up the deploy work until the hook work is completed
-		if addonHasFinalizer(addon, addonapiv1alpha1.AddonHostingPreDeleteHookFinalizer) {
+	// Don't skip syncing if the addon is deleting and there is a predelete hook, since the deployment manifests may
+	// need to be updated during the uninstall.
+	if !addonHasFinalizer(addon, addonapiv1alpha1.AddonHostingPreDeleteHookFinalizer) {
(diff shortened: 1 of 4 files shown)


zhiweiyin318:
/LGTM
the manifestworks should not be re-created when cluster/addon is deleting,  it is ok for me in this pre-delete case.
cc @qiujian16  @zhujian7

pkg/addonmanager/controllers/agentdeploy/hosted_sync.go
```diff
@@ -80,37 +80,36 @@ func (s *hostedSyncer) sync(ctx context.Context,
...
+			}
+			addonRemoveFinalizer(addon, addonapiv1alpha1.AddonHostingManifestFinalizer)
 			return addon, nil
 		}
 
-		if err = s.cleanupDeployWork(ctx, addon); err != nil {
-			return addon, err
+		// waiting for the addon to be deleted when cluster is deleting.
+		// TODO: consider to delete addon in this scenario.
+		if !cluster.DeletionTimestamp.IsZero() && !addonHasFinalizer(addon, addonapiv1alpha1.AddonHostingPreDeleteHookFinalizer) {
```
zhujian7:
```suggestion
		if !cluster.DeletionTimestamp.IsZero() {
```

mprahl: @zhujian7 I wasn't sure about this because I could see the cluster being deleted, which causes the addon to be uninstalled, and you may still want to deploy new changes to the deployment.

In the policy use-case, we want to set an annotation on the `Deployment` object to indicate it is being uninstalled and be able to deploy new image updates if the uninstall is stuck and requires a bug fix. I think we would still need to wait for the `AddonHostingPreDeleteHookFinalizer` finalizer to be removed for this to still work when the whole cluster is being uninstalled scenario.

zhujian7: yes, thanks @mprahl for the explanation.
I don't think we need to check `!addonHasFinalizer(addon, addonapiv1alpha1.AddonHostingPreDeleteHookFinalizer)` again, since we have checked this in [line 85](https://github.com/open-cluster-management-io/addon-framework/pull/226/files#diff-e4f38c04c06c31582696236d2246c4bf5d7543c79d1030ab466c42b466222117R85)

mprahl: Ah, that was a copy-paste error during development. Thanks for catching that!

mprahl:
@zhiweiyin318 could you please take another look?

zhiweiyin318:
/LGTM
need approve from @qiujian16

PR #228, opened 2023-12-14
fix e2e agent ns is failed to created when it is deleting

zhiweiyin318:


test/e2e/helloworld_helm_test.go
```diff
@@ -122,23 +122,22 @@ var _ = ginkgo.Describe("install/uninstall helloworld helm addons", func() {
...
+			err = hubKubeClient.CoreV1().Namespaces().Delete(context.Background(),
+				agentInstallNamespaceConfig, metav1.DeleteOptions{})
+			if err != nil {
+				return false
 			}
 
-			return err
-		}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
+			return false
+		}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
```
zhujian7: return error is helpful for debugging when it fails since it will print the error message, but return bool will not.

zhiweiyin318: ok, reverted.

PR #229, opened 2023-12-22
🐛 Do not check addon status when the cluster availability is unknown

zhujian7:
Do not check addon status when the cluster availability is unknown, otherwise, when the cluster is unknown and the manifestwork feedback rule deployment available replica >1(was not updated after the cluster is unknown), the addon-framework will update the addon status to `True`, but the [registration-agent will update the addon status to `Unknown`](https://github.com/open-cluster-management-io/ocm/blob/9dc8f104cf51439b6bb1f738894e75aabdf5f8dc/pkg/registration/hub/addon/healthcheck_controller.go#L68-L78), which will lead to an infinite loop


zhujian7:
@qiujian16 @zhiweiyin318

PR #230, opened 2024-01-18
🌱 Add a helper func to check if a managed cluster condition changes

zhujian7:
Add a helper func to check if a managed cluster condition changes, so users can use this to make the agent deploy controller watch if the managed cluster condition changes. 
related PR: https://github.com/open-cluster-management-io/addon-framework/pull/199

### DIFF
--- a/pkg/utils/helpers.go
+++ b/pkg/utils/helpers.go
@@ -386,8 +386,36 @@ func MapValueChanged(old, new map[string]string, key string) bool {
 // ClusterImageRegistriesAnnotationChanged returns true if the value of the ClusterImageRegistriesAnnotationKey
 // in the new managed cluster annotation is different from the old managed cluster annotation
 func ClusterImageRegistriesAnnotationChanged(old, new *clusterv1.ManagedCluster) bool {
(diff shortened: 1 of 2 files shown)


pkg/utils/helpers.go
```diff
@@ -386,8 +386,34 @@ func MapValueChanged(old, new map[string]string, key string) bool {
...
+func ClusterConditionChanged(old, new *clusterv1.ManagedCluster, conditionType string) bool {
+	if new == nil || old == nil {
+		return false
+	}
+
+	oldAvailableCondition := meta.FindStatusCondition(old.Status.Conditions, conditionType)
+	newAvailableCondition := meta.FindStatusCondition(new.Status.Conditions, conditionType)
+
+	return oldAvailableCondition != nil && newAvailableCondition != nil &&
+		oldAvailableCondition.Status != newAvailableCondition.Status
```
qiujian16: what if both condition are nil? and do we need to compare observerdGeneration and lastTransittionTime

zhujian7: if both are nil, it does not change.
if one is nil, but the other is not, it changes, I will handle this.
there is no `observerdGenereation` in the managed cluster, and we only care about the condition status, so I think `condition.status` is enough.

PR #232, opened 2024-01-22
use patcher in sdk-go

qiujian16:


zhujian7:
please also bump the golang version in `build/Dockerfile.example`

PR #236, opened 2024-01-31
add annotation if addon uses InstallStrategy

haoqing0110:
Ref: https://github.com/open-cluster-management-io/ocm/issues/355

### DIFF
--- a/examples/deploy/addon/helloworld-helm/resources/cluster_role.yaml
+++ b/examples/deploy/addon/helloworld-helm/resources/cluster_role.yaml
@@ -38,7 +38,7 @@
       verbs: ["update", "patch"]
     - apiGroups: ["addon.open-cluster-management.io"]
       resources: ["clustermanagementaddons"]
-      verbs: ["get", "list", "watch"]
+      verbs: ["get", "list", "watch", "patch"]
     - apiGroups: ["addon.open-cluster-management.io"]
       resources: ["managedclusteraddons"]
       verbs: ["get", "list", "watch", "create", "update", "delete"]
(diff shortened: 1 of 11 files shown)


pkg/addonmanager/controllers/addoninstall/controller.go
```diff
@@ -98,6 +108,27 @@ func (c *addonInstallController) sync(ctx context.Context, syncCtx factory.SyncC
 			continue
 		}
 
+		cma, err := c.clusterManagementAddonLister.Get(addonName)
```
qiujian16: this will be called per mca/cluster event. It might bring a lot of conflict. How about we put this in a controller that reconcile clustermanagementaddon only?

haoqing0110: Yes, that would be a better choice. The current code will have conflicts when the first time adding the annotation, then won't trigger patch annotation action.

qiujian16: we have another existing controller that is reconciling the clustermanageraddon

examples/deploy/addon/helloworld-helm/resources/cluster_role.yaml
```diff
@@ -38,7 +38,7 @@
       verbs: ["update", "patch"]
     - apiGroups: ["addon.open-cluster-management.io"]
       resources: ["clustermanagementaddons"]
-      verbs: ["get", "list", "watch"]
+      verbs: ["get", "list", "watch", "patch"]
```
qiujian16, line 41: note, we will need users to update the clusterrole when upgrading the library.

qiujian16 (commented):
/approve

examples/deploy/addon/helloworld-helm/resources/cluster_role.yaml
zhujian7: should be reflected in the release notes.

zhujian7:
LGTM

PR #239, opened 2024-02-04
🌱 Update addon available condition message

zhujian7:
Update addon available condition message to align with addons that [check health by lease](https://github.com/open-cluster-management-io/ocm/blob/98ab144e4216448b2e35ca2d9cdb84196f08b849/pkg/registration/spoke/addon/lease_controller.go#L145)

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
+++ b/pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
@@ -192,7 +192,7 @@ func (s *healthCheckSyncer) probeAddonStatusByWorks(
 		Type:    addonapiv1alpha1.ManagedClusterAddOnConditionAvailable,
 		Status:  metav1.ConditionTrue,
 		Reason:  addonapiv1alpha1.AddonAvailableReasonProbeAvailable,
-		Message: "Addon is available",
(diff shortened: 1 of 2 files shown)


pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
```diff
@@ -192,7 +192,7 @@ func (s *healthCheckSyncer) probeAddonStatusByWorks(
 		Type:    addonapiv1alpha1.ManagedClusterAddOnConditionAvailable,
 		Status:  metav1.ConditionTrue,
 		Reason:  addonapiv1alpha1.AddonAvailableReasonProbeAvailable,
-		Message: "Addon is available",
+		Message: fmt.Sprintf("%s add-on is available.", addon.Name),
```
qiujian16, line 195: why do we need to add the name? it is the condition of mca, the mca has the name already, isn't it?

zhujian7: this is to align with addons that [check health by lease](https://github.com/open-cluster-management-io/ocm/blob/98ab144e4216448b2e35ca2d9cdb84196f08b849/pkg/registration/spoke/addon/lease_controller.go#L145)

Issue #237, opened 2024-02-01
helm addon: GetSpecHash unable to handle  ConfigMap and Secret resources


JoaoBraveCoding:
**Describe the bug**

As far as I understand the function `GetSpecHash` is used to generate a hash with the `Spec` content of a k8s resource. This hash is stored in the status of ManagedClusterAddOn and used by the framework to know if the current state of ManifestWorks is up to date. The problem is that when we use as config resources k8s resources that don't have a `Spec` field such as ConfigMaps or Secrets.

**To Reproduce**

Deploy a `ManagedClusterAddOn` that has in its config a ConfigMap or Secret

**Expected behavior**

The framework should not discriminate these two resources and it should handle them properly

**Additional context**

Problem is with https://github.com/open-cluster-management-io/addon-framework/blob/71f1b13cbb6356279d5ae49667324ef5f9b7c3e0/pkg/utils/helpers.go#L354-L371

haoqing0110:
Similar as https://github.com/open-cluster-management-io/ocm/issues/364, will fix it.

qiujian16:
/close

this has been fixed

PR #243, opened 2024-03-06
:bug: allow arbitrary config fields when calculating config hash

### LINKED ISSUE #364: spechash remains empty for secrets
**Describe the bug**
https://kubernetes.slack.com/archives/C01GE7YSUUF/p1709566835547519

**To Reproduce**

**Expected behavior**

**Environment ie: OCM version, Kubernetes version and provider:**

**Additional context**

haoqing0110:
Fix https://github.com/open-cluster-management-io/ocm/issues/364 https://github.com/open-cluster-management-io/addon-framework/issues/237

### DIFF
--- a/pkg/utils/helpers.go
+++ b/pkg/utils/helpers.go
@@ -350,14 +350,15 @@ func IsOwnedByCMA(addon *addonapiv1alpha1.ManagedClusterAddOn) bool {
 	return false
 }
 
-// GetSpecHash returns the sha256 hash of the spec field of the given object
+// GetSpecHash returns the sha256 hash of the spec field or data field of the given object
 func GetSpecHash(obj *unstructured.Unstructured) (string, error) {
 	if obj == nil {
 		return "", fmt.Errorf("object is nil")
 	}
 	spec, ok := obj.Object["spec"]
 	if !ok {
-		return "", fmt.Errorf("object has no spec field")
+		// handle cases like secret and configmaps
+		spec = obj.Object["data"]
 	}
 
 	specBytes, err := json.Marshal(spec)
--- a/pkg/utils/helpers.go
+++ b/pkg/utils/helpers.go
@@ -350,23 +350,30 @@ func IsOwnedByCMA(addon *addonapiv1alpha1.ManagedClusterAddOn) bool {
 	return false
 }
 
-// GetSpecHash returns the sha256 hash of the spec field or data field of the given object
+// GetSpecHash returns the sha256 hash of the spec field or other config fields of the given object
 func GetSpecHash(obj *unstructured.Unstructured) (string, error) {
 	if obj == nil {
 		return "", fmt.Errorf("object is nil")
 	}
-	spec, ok := obj.Object["spec"]
-	if !ok {
-		// handle cases like secret and configmaps
-		spec = obj.Object["data"]
+
+	// create a map for config fields
+	configObj := map[string]interface{}{}
+	for k, v := range obj.Object {
+		switch k {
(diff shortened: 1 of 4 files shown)


pkg/utils/helpers.go
```diff
@@ -350,14 +350,15 @@ func IsOwnedByCMA(addon *addonapiv1alpha1.ManagedClusterAddOn) bool {
...
+// GetSpecHash returns the sha256 hash of the spec field or data field of the given object
 func GetSpecHash(obj *unstructured.Unstructured) (string, error) {
 	if obj == nil {
 		return "", fmt.Errorf("object is nil")
 	}
 	spec, ok := obj.Object["spec"]
 	if !ok {
-		return "", fmt.Errorf("object has no spec field")
+		// handle cases like secret and configmaps
+		spec = obj.Object["data"]
```
qiujian16: what if an API that does not have data field and spec field?

haoqing0110: var spec will be nil and return a hash like [case](https://github.com/open-cluster-management-io/addon-framework/pull/243/files#diff-e5c8046fa2e55ccd0b73a7f6cd75cfd2798f8e7bec08210444e01d6ffac0c0f0R91) .For example, if a configmap removed data field, hash will update and trigger an upgrade.

qiujian16: how about we remove status and metadata part and hash?

haoqing0110: Yes, sounds better. Have modified the code.

pkg/utils/helpers.go
```diff
@@ -350,22 +350,25 @@ func IsOwnedByCMA(addon *addonapiv1alpha1.ManagedClusterAddOn) bool {
...
+	delete(obj.Object, "status")
+
+	// the object key will also be included in the hash calculation to ensure that different fields with the same value have different hashes
+	configBytes, err := json.Marshal(obj.Object)
 	if err != nil {
 		return "", err
 	}
 
-	hash := sha256.Sum256(specBytes)
+	hash := sha256.Sum256(configBytes)
```
tamalsaha, line 376: May be a fast non-cryptographic is good enough here?
https://github.com/zeebo/xxh3

pkg/utils/helpers.go
```diff
@@ -350,22 +350,25 @@ func IsOwnedByCMA(addon *addonapiv1alpha1.ManagedClusterAddOn) bool {
...
 		return "", fmt.Errorf("object is nil")
 	}
-	spec, ok := obj.Object["spec"]
-	if !ok {
-		return "", fmt.Errorf("object has no spec field")
-	}
 
-	specBytes, err := json.Marshal(spec)
+	// remove non config related fields, the remaining filed should be spec, data, other use defined fields or even empty
+	delete(obj.Object, "apiVersion")
```
tamalsaha: This will modify the input object. Is that going to cause any unintended issue? May be a better choice is to perform a shallow copy into a new map and use that calculate hash?

```
cp := map[string]any{}
for k, v := range obj.Object {
  switch k: {
    case "apiVersion", "kind", "metadata", "status":
    default:
       cp[k] = v
   }
} 
```

haoqing0110: Thanks for raising this, have modified the code.

pkg/utils/helpers.go
haoqing0110: This seems to have better performance, but might not be suitable for the addon case, for example, a cluster-scoped config hash will be exposed to cma, mca and work, and might be cracked by an unexpected namespace-level user.

PR #245, opened 2024-03-07
:bug: fix deploying ocm error in e2e

zhujian7:
The [PR in ocm repo](https://github.com/open-cluster-management-io/ocm/pull/365/files#diff-191c20bda57b28bc09dad6d2a883b7431ebf2aaae26c53828ca91bea09d25fa2R7) changed the default install mode to `Singleton`, so there will be only one `klusterlet-agent` on the manged cluster


haoqing0110 (approved):
/lgtm

PR #246, opened 2024-03-11
🌱 Add a method to set the hosting cluster for the Helm agent

mprahl:
The policy addons require setting this since we need the hosting cluster capabilities variables and we can't guarantee that the hosting cluster is static for each addon.

### DIFF
--- a/pkg/addonfactory/helm_agentaddon.go
+++ b/pkg/addonfactory/helm_agentaddon.go
@@ -219,6 +219,10 @@ func (a *HelmAgentAddon) getBuiltinValues(
 	return helmBuiltinValues, nil
 }
 
+func (a *HelmAgentAddon) SetHostingCluster(hostingCluster *clusterv1.ManagedCluster) {
+	a.hostingCluster = hostingCluster
+}
+
 func (a *HelmAgentAddon) getDefaultValues(
 	cluster *clusterv1.ManagedCluster,
 	addon *addonapiv1alpha1.ManagedClusterAddOn) (Values, error) {


pkg/addonfactory/helm_agentaddon.go
```diff
@@ -219,6 +219,10 @@ func (a *HelmAgentAddon) getBuiltinValues(
 	return helmBuiltinValues, nil
 }
 
+func (a *HelmAgentAddon) SetHostingCluster(hostingCluster *clusterv1.ManagedCluster) {
```
zhujian7, line 222: Hi @mprahl, there is a [WithHostingCluster](https://github.com/open-cluster-management-io/addon-framework/blob/974d3377ecbd1c74fecb801148c10b4664556be5/pkg/addonfactory/addonfactory.go#L126) method for agentAddonFactory already, not sure if that is what you need?

mprahl: @zhujian7 the issue is that `WithHostingCluster` forces a single value for all hosted clusters, however, the hosting cluster is set at the `ManagedClusterAddOn` object level, so it could be different hosting clusters for different hosted clusters for the same addon.

zhujian7: Yes, you are right, but I am still confused about how the func `SetHostingCluster` to the `HelmAgentAddon` can set different hosting clusters for different hosted clusters for the same addon,  could you give some examples/tests of how to use the added `SetHostingCluster` for addon developers?

mprahl: It's for this section of code here:
https://github.com/open-cluster-management-io/governance-policy-addon-controller/blob/46f8f4a9de0b6854f362981389b71bcff7d7a74f/pkg/addon/common.go#L175-L197

zhujian7: emmm... but setting hostingCluster in the `Manifests` method is weird.
@zhiweiyin318 @qiujian16 @haoqing0110 WDYT? any suggested way to implement this?

mprahl: @zhujian7 I'm open to other ideas but in #149, the maintainers said that changing the method signatures by adding additional arguments was a no go.

Another option is to have `Manifests` in the addon framework query for the hosting cluster if set on the addon like we do in our code rather than have it as a struct field.

zhujian7: I am think: can we remove the [`hostingCluster` field for struct `HelmAgentAddon`](https://github.com/open-cluster-management-io/addon-framework/blob/d22ba0c3b84cedc510d428ee94d75cd84008af98/pkg/addonfactory/helm_agentaddon.go#L52), and not use `if a.hostingCluster != nil` [here](https://github.com/open-cluster-management-io/addon-framework/blob/d22ba0c3b84cedc510d428ee94d75cd84008af98/pkg/addonfactory/helm_agentaddon.go#L259), instead we get the hostingcluster name from `managedClusterAddon.annotations["addon.open-cluster-management.io/hosting-cluster-name"]`,  and then use a hub clusterCient to get the hostingcluster.
The main reason I think this is better is: from my point of view the `HelmAgentAddon` is used for config the addon for all managed clusters(it is like per clustermanagementaddon), but the hostingcluster is for one managed cluster(per managedclusteraddon), so it should not be a field of the `HelmAgentAddon`

mprahl: @zhujian7 , yes, I agree but I don't have the capacity to do this refactoring right now. If you do it, I'd be happy to adopt that change.

zhujian7: Hi @mprahl I submitted a PR https://github.com/open-cluster-management-io/addon-framework/pull/251, please help take a review.

PR #249, opened 2024-03-11
:bug: Sort manifests for by Kind helm addons

tamalsaha:
Helm will sort objects by kind when applying to a cluster. This pr also follows/copies the same code to sort rendered objects.


pkg/addonfactory/helm_agentaddon.go
```diff
@@ -260,3 +279,47 @@ func (a *HelmAgentAddon) releaseOptions(
 		Namespace: a.getValueAgentInstallNamespace(addon),
 	}
 }
+
+// Manifest represents a manifest file, which has a name and some content.
+type Manifest struct {
```
qiujian16: nit, should be private struct. I think you only use it internally?

tamalsaha: Done!

PR #252, opened 2024-03-19
:seedling: remove duplicate controller and installstrategy

haoqing0110:
## Summary

Ref https://github.com/open-cluster-management-io/ocm/issues/355
- Remove duplicate config controller and owner controller code in framework.
- Remove [WithInstallStrategy()](https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/addonfactory/addonfactory.go#L84) func in framework.
- If use addon-framework, remove annotation addon.open-cluster-management.io/lifecycle: self.

## Related issue(s)

Fixes #


pkg/addonmanager/controllers/cmamanagedby/controller.go
```diff
@@ -17,11 +17,15 @@ import (
 )
 
 const (
-	controllerName = "management-addon-controller"
+	controllerName = "cma-managed-by-controller"
 )
 
-// clusterManagementAddonController reconciles cma on the hub.
-type clusterManagementAddonController struct {
+// cmaManagedByController reconciles clustermanagementaddon on the hub
```
qiujian16, line 23: Is it necessary to remove this? will remove this controller later when addon manager no longer honor this annotation?

haoqing0110: Keep the controller is to remove the "self" added by framework 0.8.1/0.9.0. 

> will remove this controller later when addon manager no longer honor this annotation?

Yes. May need to keep this as long as we support 0.8.1/0.9.0, or users have to remove the "self" manually when they do a framework upgrade.

PR #254, opened 2024-03-20
🌱 Start ocm env using clusteradm

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/examples/deploy/hosted-ocm/install.sh
+++ b/examples/deploy/hosted-ocm/install.sh
@@ -20,16 +20,6 @@ function wait_deployment() {
   set -e
 }
 
-function hub_approve_cluster(){
-  local cluster_name=$1
-  echo "approve ${cluster_name}"
-  ${KUBECTL} get csr -l open-cluster-management.io/cluster-name=${cluster_name} | \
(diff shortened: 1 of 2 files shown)


zhujian7:
@dhaiducek @mprahl FYI: I know there seems to be a policy repo that uses the cloning ocm repo and executing make commands to install an ocm env, which is error-prone, here is an example of using the clusteradm to set up an ocm env for your reference, hope this is helpful.

dhaiducek:
@zhujian7 Thanks for tagging us! `clusteradm` could be a good solution and we'd definitely be on board for a more reliable build, but for our use-cases:
- We want to be able to deploy `stolostron`-specific code/images from a particular version (which I'm not sure would be quite as simple with `clusteradm`?)
- We want test against the latest images rather than a specific version

Would both of these be straightforward with `clusteradm` also?

zhujian7:
@dhaiducek I am not quite sure if this is what you need, but you can have a try.
there are `--bundle-version='xx'` and `--image-registry='xxx'` options for the `clusteradm init` and `clusteradm join` commands.

PR #256, opened 2024-03-26
✨ support config work driver.

morvencao:
## Summary

This PR aims to enhance the addon-manager by adding work driver configuration support.
This enables the addon manager to deploy addon agents using drivers other than `kube`.

## Related issue(s)

ref: https://issues.redhat.com/browse/ACM-10423


go.mod
```diff
@@ -5,24 +5,27 @@ go 1.21
...
-	k8s.io/client-go v0.29.1
+	k8s.io/client-go v0.29.2
 	k8s.io/component-base v0.29.1
-	k8s.io/klog/v2 v2.110.1
-	k8s.io/utils v0.0.0-20240102154912-e7106e64919e
+	k8s.io/klog/v2 v2.120.1
+	k8s.io/utils v0.0.0-20240310230437-4693a0247e57
 	open-cluster-management.io/api v0.13.0
-	open-cluster-management.io/sdk-go v0.13.0
+	open-cluster-management.io/ocm v0.13.1-0.20240325044258-22501d88f774
```
qiujian16: we should avoid import ocm

morvencao: removed ocm dependency.

pkg/addonmanager/manager.go
```diff
@@ -321,10 +345,40 @@ func (a *addonManager) StartWithInformers(ctx context.Context,
...
+	flags.StringVar(&o.WorkDriverConfig, "work-driver-config",
+		o.WorkDriverConfig, "The config file path of current work driver")
+	flags.StringVar(&o.CloudEventsClientID, "cloudevents-client-id",
+		o.CloudEventsClientID, "The ID of the cloudevents client when publishing works with cloudevents")
+	flags.StringVar(&o.SourceID, "source-id",
+		o.SourceID, "The ID of the source when publishing works with cloudevents")
+}
+
+// New returns a new addon manager for creating addon agents.
+func New(config *rest.Config, opts *ManagerOptions) (AddonManager, error) {
```
qiujian16: this will break the existing signature of the function. the opts filed should be an optional field. And we should allow that it is not set.

cmd/example/helloworld/main.go
```diff
@@ -66,22 +66,30 @@ func newCommand() *cobra.Command {
 }
 
 func newControllerCommand() *cobra.Command {
+	o := addonmanager.NewManagerOptions()
```
qiujian16: we should not impact the existing addon. We could build a new addon example that supports cloudevent, or select one addon as an example. It also ensure that existing addon won't be impacted by this change.

qiujian16 (commented):
can we enable the integration for cloudevents？

test/integration-test.mk
```diff
@@ -26,7 +26,15 @@ clean-integration-test:
...
+	go test -c ./test/integration/kube -o ./kube-integration.test
+	./kube-integration.test -ginkgo.slowSpecThreshold=15 -ginkgo.v -ginkgo.failFast
+.PHONY: test-kube-integration
+
+test-cloudevents-integration: ensure-kubebuilder-tools
+	go test -c ./test/integration/cloudevents -o ./cloudevents-integration.test
+	./cloudevents-integration.test -ginkgo.slowSpecThreshold=15 -ginkgo.v -ginkgo.failFast
+.PHONY: test-cloudevents-integration
+
+test-integration: test-kube-integration test-cloudevents-integration
```
morvencao: created separate folder for cloudevents integration test.
make sense? @qiujian16 @skeeey

pkg/addonmanager/manager.go
morvencao: updated this to make managerOptions be optional.

cmd/example/helloworld/main.go
morvencao: built new example that supports cloudevent, following the existing examples.

test/integration-test.mk
skeeey: yeah, we may do some configuration like this https://github.com/open-cluster-management-io/ocm/blob/main/.github/workflows/cloudevents-integration.yml#L5

pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -90,6 +92,9 @@ func newManifestWork(addonNamespace, addonName, clusterName string, manifests []
 			Labels: map[string]string{
 				addonapiv1alpha1.AddonLabelKey: addonName,
 			},
+			Annotations: map[string]string{
+				common.CloudEventsDataTypeAnnotationKey: payload.ManifestBundleEventDataType.String(),
```
qiujian16: We need to add this annotation all the time? I do not think we should add any additionaly annotation to existing manifestwork.

pkg/addonmanager/manager.go
```diff
@@ -321,12 +345,50 @@ func (a *addonManager) StartWithInformers(ctx context.Context,
 	return nil
 }
 
-// New returns a new Manager for creating addon agents.
-func New(config *rest.Config) (AddonManager, error) {
-	return &addonManager{
+// ManagerOptions defines the flags for addon manager
```
qiujian16: can we put this Option flag in the separated dir specifically for cloudevent?

pkg/addonmanager/manager.go
```diff
@@ -81,12 +87,50 @@ func (a *addonManager) Trigger(clusterName, addonName string) {
...
+	var addonNames []string
+	for key := range a.addonAgents {
+		addonNames = append(addonNames, key)
+	}
+
+	// To support sending ManifestWorks to different drivers (like the Kubernetes apiserver or MQTT broker), we build
+	// ManifestWork client that implements the ManifestWorkInterface and ManifestWork informer based on different
+	// driver configuration.
+	// Refer to Event Based Manifestwork proposal in enhancements repo to get more details.
+	_, config, err := cloudeventswork.NewConfigLoader(a.options.WorkDriver, a.options.WorkDriverConfig).
```
qiujian16: could we move all this cloudEvent setting to a separated func instead of the main Start func? or a separated cloudEvent manager? I am concerned about this change on the existing addon-framework breaking existing addons.

pkg/addonmanager/controllers/agentdeploy/utils.go
morvencao: yes, we do need this all the time. /cc @skeeey

skeeey: we need this annotation to indicate the manifest data type, but I think we may be able to enhance the sdk-go, if we don't find this annotation, we use `ManifestBundle` as a  default value?

skeeey: and for ocm/acm cases, it should be always `ManifestBundle`

morvencao: yes, we can set default value(bundle type) in sdk-go so that we don't need set it here explicitly.

pkg/addonmanager/manager.go
morvencao: put this to separate cloudevents package.

pkg/addonmanager/manager.go
morvencao: Implemented `cloudeventsAddonManager` in separate package `cloudevents`, so that we don't break existing addons. see: https://github.com/open-cluster-management-io/addon-framework/pull/256/commits/dcf9539ec9090eb372454e2c626fe4528364fce7

pkg/addonmanager/manager.go
```diff
@@ -47,12 +47,12 @@ type AddonManager interface {
...
-		workInformers workv1informers.SharedInformerFactory,
-		addonInformers addoninformers.SharedInformerFactory,
-		clusterInformers clusterv1informers.SharedInformerFactory,
-		dynamicInformers dynamicinformer.DynamicSharedInformerFactory) error
+	// StartWithInformers(ctx context.Context,
+	// 	kubeInformers kubeinformers.SharedInformerFactory,
+	// 	workInformers workv1informers.SharedInformerFactory,
+	// 	addonInformers addoninformers.SharedInformerFactory,
+	// 	clusterInformers clusterv1informers.SharedInformerFactory,
+	// 	dynamicInformers dynamicinformer.DynamicSharedInformerFactory) error
```
morvencao: @qiujian16 @skeeey 
The only concern I have, is it safe to remove this? I didn't find any place the function is called externally.
I need to remove this or change the function siginature because in the `cloudeventsAddonManager` implementation, we can't get the `workv1informers.SharedInformerFactory` from clientHolder (from sdk-go), we can only get `workv1informers.Informer` type.

pkg/addonmanager/controllers/agentdeploy/utils.go
skeeey: or we can give a fixed value(bundle type) now

pkg/addonmanager/manager.go
qiujian16: I think we use this ocm for addon Template

qiujian16 (commented):
let's remove the whole e2e part, I think we decide not to include e2e in this PR.

qiujian16:
@zhujian7 @zhiweiyin318 pls take a careful look at this one, to avoid any compatibility issue

pkg/addonmanager/controllers/agentdeploy/utils.go
morvencao: updated sdk-go dependency, removed this annotation.

pkg/addonmanager/manager.go
morvencao: I reverted the change in `AddonManager` interface, added a mock (TODO item) implementation in the `cloudeventsAddonManager ` implementation. see: https://github.com/open-cluster-management-io/addon-framework/pull/256/commits/5c1a6e7a994360ea141655c3e8eb3bf1cf9379ed

morvencao:
kindly ping @zhujian7 @zhiweiyin318 ~

zhiweiyin318:
Look good to me.

pkg/addonmanager/cloudevents/manager.go
```diff
@@ -0,0 +1,348 @@
...
+	workInformers workinformers.SharedInformerFactory,
+	addonInformers addoninformers.SharedInformerFactory,
+	clusterInformers clusterv1informers.SharedInformerFactory,
+	dynamicInformers dynamicinformer.DynamicSharedInformerFactory) error {
+	// TODO: Implement this method to support usage within the addon template.
+	// Requires updating sdk-go to enable clientHolder to manage the shared informer for ManifestWork.
+	return fmt.Errorf("method StartWithInformers is not implemented")
+}
+
+func (a *cloudeventsAddonManager) startWithInformers(ctx context.Context,
```
qiujian16: actully, how about we have a baseManager interface with StartWithInformers method and an implementation that new all controllers. The Manager interface inlines the baseManager interface, and the manager/cloudeventmanger have different implementation of start func, but share the same implementation in baseManager. Will that work?

morvencao: Updated this, not sure if this is more clear.

qiujian16: why we still need this? I thought this should be in the baseManager.

morvencao: removed this, PTAL @qiujian16

examples/deploy/ocm-cloudevents/install.sh
```diff
@@ -0,0 +1,87 @@
...
+stringData:
+  config.yaml: |
+    brokerHost: mosquitto.mqtt:1883
+    topics:
+      sourceEvents: sources/addon-manager/clusters/${MANAGED_CLUSTER_NAME}/sourceevents
+      agentEvents: sources/addon-manager/clusters/${MANAGED_CLUSTER_NAME}/agentevents
+      agentBroadcast: clusters/${MANAGED_CLUSTER_NAME}/agentbroadcast
+EOF
+
+# patch klusterlet-work-agent deployment to use mqtt as workload source driver
```
zhujian7, line 54: Can the klusterlet-work-agent be deployed in the way that the mqtt is the driver directly?

examples/helloworld_cloudevents/helloworld.go
```diff
@@ -0,0 +1,95 @@
...
+	}{
+		KubeConfigSecret: fmt.Sprintf("%s-hub-kubeconfig", addon.Name),
+		ClusterName:      cluster.Name,
+		Image:            image,
+	}
+
+	return addonfactory.StructToValues(manifestConfig), nil
+}
+
+func AgentHealthProber() *agent.HealthProber {
```
zhujian7: is there any specific reason that we need to use the work prober to check the health for cloud events driver addons?

pkg/addonmanager/manager.go
```diff
@@ -189,144 +137,9 @@ func (a *addonManager) Start(ctx context.Context) error {
...
-		go csrApproveController.Run(ctx, 1)
-	}
-	if csrSignController != nil {
-		go csrSignController.Run(ctx, 1)
-	}
-	return nil
-}
-
 // New returns a new Manager for creating addon agents.
 func New(config *rest.Config) (AddonManager, error) {
```
zhujian7, line 141: Not sure if it is better to provider a addon manager factory func, like:
```
func New(mtype AddonManagerType)(AddonManager, error){
  if mtype =="cloudevents"{
    return ...
  }
  ...
}
```

pkg/addonmanager/cloudevents/manager.go
```diff
@@ -0,0 +1,289 @@
...
+	"open-cluster-management.io/addon-framework/pkg/basecontroller/factory"
+	"open-cluster-management.io/addon-framework/pkg/index"
+	"open-cluster-management.io/addon-framework/pkg/utils"
+	cloudeventswork "open-cluster-management.io/sdk-go/pkg/cloudevents/work"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/work/source/codec"
+)
+
+// cloudeventsAddonManager is the implementation of AddonManager with
+// the base implementation and cloudevents options
+type cloudeventsAddonManager struct {
```
zhujian7, line 29: could you explain the difference between the cloudevents addon manager and the classic addon manager except for the building way of manifestwork client and informer? which I think is good for review.

.github/workflows/cloudevents-integration.yml
```diff
@@ -0,0 +1,32 @@
...
+
+env:
+  GO_VERSION: '1.21'
+  GO_REQUIRED_MIN_VERSION: ''
+
+permissions:
+  contents: read
+
+jobs:
+  integration:
```
zhujian7, line 21: can we also add an e2e to at least check that the example cloud event addons are available when deploying into a kubernetes cluster?

examples/deploy/ocm-cloudevents/install.sh
morvencao: Yes, the latest work agent image already support mqtt and grpc drivers.

examples/helloworld_cloudevents/helloworld.go
morvencao: the cloudevents driver added in this PR is only used for addon agent deploy, no other difference from the current "kube" driver. Therefore, it use similar prober to make sure addon agent is deployed successfully on managed cluster as "kube" driver. Make sense?

pkg/addonmanager/manager.go
morvencao: this will break existing code that using `addonmanager.New(Config)` to create addon manager, right?
we don't want make any change to current code, that's why we keep this unchanged.

pkg/addonmanager/cloudevents/manager.go
morvencao: the difference between the cloudevents addon manager and the classic addon manager is the way to build work client and work informer.

the cloudevents addon manager use cloudevents drivers(mqtt or grpc) to create work informer, so that the manifestwork used to deploy addon agent can be delivered to managed cluster through mqtt or grpc.

no other difference.

.github/workflows/cloudevents-integration.yml
morvencao: e2e test was removed in this commit: https://github.com/open-cluster-management-io/addon-framework/pull/256/commits/d8d4f5269b42f4e24b4917ce38d372da62aa56a6
because the work client based on cloudevents drivers (sdk-go) are missing GC machnism, we will revisit and add e2e after GC machnism support for cloudevents drivers.

pkg/addonmanager/cloudevents/manager.go
qiujian16: I mean all the code to init controllers and start them. I think they are duplicated code and should exist only in baseManager. If we have any change on controllers in the future, we do not need to update multiple times.

morvencao: We may not do that, due to the limitation source work client from sdk-go as I mentioned at https://github.com/open-cluster-management-io/addon-framework/pull/256#discussion_r1566694777

we can't get the `workinformers.SharedInformerFactory` from `clientHolder` (from sdk-go), we can only get `workv1informers.Informer` type, so we can't call `StartWithInformers` to start the controllers in cloudevents addon manager.

morvencao:
@qiujian16 @zhujian7 Another look?

pkg/addonmanager/cloudevents/manager.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+			listOptions.LabelSelector = metav1.FormatLabelSelector(selector)
+		}),
+	)
+
+	workClient := clientHolder.WorkInterface()
+	workInformers := clientHolder.ManifestWorkInformer()
+
+	// addonDeployController
+	err = workInformers.Informer().AddIndexers(
```
qiujian16, line 124: nit: the indexer part should be put in a shared dir. It is duplicated with the manager.

qiujian16 (commented):
/approve

morvencao:
added back the e2e testing for helloworld addon using cloudevents.
/unhold

@zhujian7 PTAL

zhujian7:
LGTM

PR #257, opened 2024-03-28
:warning: check return err for customized agent install namespace

zhujian7:
## Summary

This PR changes the `RegistrationOption` interface:
from
```
AgentInstallNamespace func(addon *addonapiv1alpha1.ManagedClusterAddOn) string
```
to
```
AgentInstallNamespace func(addon *addonapiv1alpha1.ManagedClusterAddOn) (string, error)
```
we need this error info to determine whether the namespace is empty or an error occurred.

## Related issue(s)

Fixes #

### DIFF
--- a/examples/deploy/addon/helloworld-helm/resources/helloworld_helm_clustermanagementaddon.yaml
+++ b/examples/deploy/addon/helloworld-helm/resources/helloworld_helm_clustermanagementaddon.yaml
@@ -9,4 +9,7 @@ spec:
   supportedConfigs:
   - group: addon.open-cluster-management.io
     resource: addondeploymentconfigs
+    defaultConfig:
+      name: helloworldhelm-config
+      namespace: open-cluster-management
   - resource: configmaps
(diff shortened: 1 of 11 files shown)


examples/deploy/addon/helloworld-helm/resources/helloworld_helm_clustermanagementaddon.yaml
```diff
@@ -9,4 +9,7 @@ spec:
   supportedConfigs:
   - group: addon.open-cluster-management.io
     resource: addondeploymentconfigs
+    defaultConfig:
```
qiujian16: this is weird, if we do to specify the default configs, we should just use the default ns, right?

zhujian7: If we do not specify the default config, we must specify it in every e2e case.

qiujian16 (commented):
/approve
/lgtm

PR #260, opened 2024-04-11
:bug:  Set a finalizer on ClusterManagementAddOn objects

yiraeChristineKim:
When GRC is disabled globally within ACM, all components, including the Policy Addon controller get deleted. This means that any finalizer cleanup the addon controller would normally do, doesn't have a chance to finish.

The proposal is for the Policy Addon controller to add finalizers to the ClusterManagementAddOn objects associated with the policy addons when one or more managed cluster has the addon installed. The finalizer is removed when the policy addons are successfully cleaned up on all managed clusters.

In addition, when the ACM installer (or something else) deletes the ClusterManagementAddOn, the Policy Addon controller should start uninstalling the associated addon. The finalizer removal would work as described above.

Ref: https://issues.redhat.com/browse/ACM-8473

## Related issue(s)

Fixes #


zhujian7:
@qiujian16 what do you think about this change?

zhujian7:
I am not sure if this is needed, because I think all managedclusteraddon CRs have an `ownerReferences` to the clustermanagementaddon, when the clustermanagementaddon is deleted, all the mangedclusteraddon will be deleted.

yiraeChristineKim:
> I am not sure if this is needed, because I think all managedclusteraddon CRs have an `ownerReferences` to the clustermanagementaddon, when the clustermanagementaddon is deleted, all the mangedclusteraddon will be deleted.

You're right. We are having a conversation with the hub installer team. https://github.com/open-cluster-management-io/addon-framework/pull/260#issuecomment-2050855825. After the conversation is over, then I will open this PR

pkg/addonmanager/controllers/cmamanagedby/controller.go
```diff
@@ -92,3 +100,42 @@ func (c *cmaManagedByController) sync(ctx context.Context, syncCtx factory.SyncC
...
+				if err != nil {
+					return err
+				}
+			}
+
+			// Removing addon, manifestWork and managedClusterAddon takes time
+			time.Sleep(1000)
+			// Error for resync
+			return fmt.Errorf("Wait for all associated managedClusterAddon to be deleted")
+		}
```
mprahl: Do we need this since there are owner references on the managed cluster addons of the CMA and CMA is deleting?

pkg/addonmanager/controllers/cmamanagedby/controller.go
```diff
@@ -92,3 +100,42 @@ func (c *cmaManagedByController) sync(ctx context.Context, syncCtx factory.SyncC
 	_, err = c.addonPatcher.PatchLabelAnnotations(ctx, cmaCopy, cmaCopy.ObjectMeta, cma.ObjectMeta)
 	return err
 }
+
+func (c *cmaManagedByController) handleFinalizer(ctx context.Context, cma *addonapiv1alpha1.ClusterManagementAddOn, addonName string) error {
```
mprahl, line 102: I think this method can be simplified to if there is one or more ManagedClusterAddOn objects, ensure the finalizer on the CMA, otherwise, remove the finalizer.

mprahl:
> I am not sure if this is needed, because I think all managedclusteraddon CRs have an `ownerReferences` to the clustermanagementaddon, when the clustermanagementaddon is deleted, all the mangedclusteraddon will be deleted.

@zhujian7 so the issue is that the hub installer deletes the ClusterManagementAddOn when uninstalling the GRC components and then deletes the governance-policy-addon-controller without waiting for the addons to be uninstalled. This is fine unless there is a predelete hook involved which requires coordination from the addon framework.

The proposal here is to have a finalizer on the ClusterManagementAddOn until all uninstalls have completed so the hub installer can watch this one object and then once it is deleted, continue with removing the governance-policy-addon-controller.

PR #266, opened 2024-05-14
:book: Update readme with addon contributions link

mikeshng:
## Summary
Update the README with addon contribution repo link

## Related issue(s)

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,6 @@
 
 This is to define an AddOn framework library.
 
-Still in PoC phase.
-
 ## Community, discussion, contribution, and support
 
 Check the [CONTRIBUTING Doc](CONTRIBUTING.md) for how to contribute to the repo.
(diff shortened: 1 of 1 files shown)


mikeshng:
Doc update like this one should not trigger e2e. 
Did I not do something in the PR?

xuezhaojun:
> Doc update like this one should not trigger e2e. Did I not do something in the PR?

No, I met the same e2e which should not occur, it seems like some thiing addon-framework e2e depends on changed, for example the images it use or the clusteradmin.

mikeshng:
> > Doc update like this one should not trigger e2e. Did I not do something in the PR?
> 
> No, I met the same e2e which should not occur, it seems like some thiing addon-framework e2e depends on changed, for example the images it use or the clusteradmin.

Thanks and go to sleep :sweat_smile:

PR #269, opened 2024-05-16
🐛 read config referent from the desired config

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/utils/addon_config.go
+++ b/pkg/utils/addon_config.go
@@ -120,7 +120,9 @@ func GetAddOnDeploymentConfigSpecHash(config *addonapiv1alpha1.AddOnDeploymentCo
 	})
 }
 
-// GetAddOnConfigRef returns the first addon config ref for the given config type
+// GetAddOnConfigRef returns the first addon config ref for the given config type. It is fine since the status will only
+// have one config for each GK.
+// (TODO) this needs to be reconcidered if we support multiple same GK in the config referencese.
 func GetAddOnConfigRef(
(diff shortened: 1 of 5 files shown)


pkg/utils/addon_config.go
```diff
@@ -125,9 +125,10 @@ func GetAddOnConfigRef(
 	configReferences []addonapiv1alpha1.ConfigReference,
```
haoqing0110, line 127: > // GetAddOnConfigRef returns the first addon config ref for the given config type


Need to update the func description as well.

pkg/utils/addon_config.go
```diff
@@ -120,14 +120,15 @@ func GetAddOnDeploymentConfigSpecHash(config *addonapiv1alpha1.AddOnDeploymentCo
...
-// GetAddOnConfigRef returns the first addon config ref for the given config type
+// GetAddOnConfigRef returns the last addon config ref for the given config type
 func GetAddOnConfigRef(
 	configReferences []addonapiv1alpha1.ConfigReference,
 	group, resource string) (bool, addonapiv1alpha1.ConfigReference) {
 
-	for _, config := range configReferences {
-		if config.Group == group && config.Resource == resource {
-			return true, config
+	// iterate the config reversely, the one with larger index is picked at first
```
zhujian7: Have we already sorted this in another controller(addonconfiguration controller, with a graph structure)? I need to confirm this.

qiujian16: this is use in the helm template addon to get values, which will need this to read in reverse order.

PR #275, opened 2024-06-13
✨  Support addon health probe for checking workload availability

zhujian7:
## Summary

Add a new health check probe type `WorkloadAvailability` to check if all deployments and daemonsets are available.

## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/368

Fixes #

### DIFF
--- a/pkg/utils/probe_helper.go
+++ b/pkg/utils/probe_helper.go
@@ -6,6 +6,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	workapiv1 "open-cluster-management.io/api/work/v1"
 
@@ -41,7 +42,7 @@ func (d *DeploymentProber) ProbeFields() []agent.ProbeField {
 	for _, deploy := range d.deployments {
 		probeFields = append(probeFields, agent.ProbeField{
 			ResourceIdentifier: workapiv1.ResourceIdentifier{
-				Group:     "apps",
+				Group:     appsv1.GroupName,
 				Resource:  "deployments",
 				Name:      deploy.Name,
 				Namespace: deploy.Namespace,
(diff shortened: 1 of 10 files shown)


pkg/utils/probe_helper.go
```diff
@@ -56,30 +57,60 @@ func (d *DeploymentProber) ProbeFields() []agent.ProbeField {
...
 	for _, value := range result.Values {
-		if value.Name != "ReadyReplicas" {
-			continue
+		if value.Name == "ReadyReplicas" {
+			readyReplicas = int(*value.Value.Integer)
 		}
+		if value.Name == "Replicas" { // for deployment
+			desiredNumberReplicas = int(*value.Value.Integer)
+		}
+		if value.Name == "DesiredNumberScheduled" { // for daemonset
```
zhujian7: Need to add a new well known status for daemonsets.

zhujian7: https://github.com/open-cluster-management-io/ocm/pull/518

pkg/utils/probe_helper.go
```diff
@@ -56,30 +57,66 @@ func (d *DeploymentProber) ProbeFields() []agent.ProbeField {
...
 	}
-	return fmt.Errorf("readyReplica is not probed")
+
+	switch identifier.Resource {
+	case "deployments":
+		if desiredNumberReplicas == 0 || readyReplicas >= 1 {
+			return nil
+		}
+	case "daemonsets":
+		if readyReplicas == desiredNumberReplicas && readyReplicas > -1 {
```
zhujian7, line 113: @qiujian16 @zhiweiyin318  Do you think this is reasonable for daemonsets?

zhujian7: If a cluster has no nodes, when deploying a daemonset, it will be like:
```
╰─$ oc get ds nginx-deployment
NAME               DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
nginx-deployment   0         0         0       0            0           <none>          18s
```
should we treat it as health or not?
as the current implementation, because the `readyReplicas(0) == desiredNumberReplicas(0)`, it is health.

qiujian16: should be healthy in this case.

PR #277, opened 2024-06-18
:seedling: refactor basecontroller with sdk-go and move certrotation to sdk-go

zhiweiyin318:
## Related issue(s)

Fixes # https://github.com/open-cluster-management-io/addon-framework/issues/276


qiujian16 (commented):
we should also remove be basecontroller or mark them as deprecated.

zhiweiyin318:
> we should also remove be basecontroller or mark them as deprecated.

@qiujian16 yes, basecontroller has been removed too.

PR #284, opened 2024-09-30
:bug: stop rendering manifests when addon configured condition is not true

haoqing0110:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/controller.go
+++ b/pkg/addonmanager/controllers/agentdeploy/controller.go
@@ -381,6 +381,12 @@ func (c *addonDeployController) buildDeployManifestWorksFunc(addonWorkBuilder *a
 			return nil, nil, fmt.Errorf("failed to get agentAddon")
 		}
 
+		if agentAddon.GetAgentAddonOptions().ConfigCheckEnabled &&
+			!meta.IsStatusConditionTrue(addon.Status.Conditions, addonapiv1alpha1.ManagedClusterAddOnConditionConfigured) {
+			klog.InfoS("Addon configured condition is not set in status", "addonName", addon.Name)
+			return nil, nil, nil
+		}
+
 		objects, err := agentAddon.Manifests(cluster, addon)
 		if err != nil {
 			meta.SetStatusCondition(&addon.Status.Conditions, metav1.Condition{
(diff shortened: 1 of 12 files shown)


pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -381,6 +381,12 @@ func (c *addonDeployController) buildDeployManifestWorksFunc(addonWorkBuilder *a
 			return nil, nil, fmt.Errorf("failed to get agentAddon")
 		}
 
+		if agentAddon.GetAgentAddonOptions().ConfigCheckEnabled &&
+			!meta.IsStatusConditionTrue(addon.Status.Conditions, addonapiv1alpha1.ManagedClusterAddOnConditionConfigured) {
+			klog.InfoS("Addon configured condition is not set in status", "addonName", addon.Name)
+			return nil, nil, fmt.Errorf("addon %s configured condition is not is not set in status", addon.Name)
```
qiujian16: do we need to return error here? I think we do not, the update of addon will retrigger the reconcile.

pkg/addonmanager/controllers/agentdeploy/controller.go
```diff
@@ -443,6 +449,12 @@ func (c *addonDeployController) buildHookManifestWorkFunc(addonWorkBuilder *addo
 			return nil, fmt.Errorf("failed to get agentAddon")
 		}
 
+		if agentAddon.GetAgentAddonOptions().ConfigCheckEnabled &&
+			!meta.IsStatusConditionTrue(addon.Status.Conditions, addonapiv1alpha1.ManagedClusterAddOnConditionConfigured) {
+			klog.InfoS("Addon configured condition is not set in status", "addonName", addon.Name)
+			return nil, fmt.Errorf("addon %s configured condition is not is not set in status", addon.Name)
```
qiujian16: ditto

pkg/addonmanager/controllers/agentdeploy/controller.go
haoqing0110: fixed.

pkg/addonmanager/controllers/agentdeploy/controller.go
haoqing0110: fixed.

Issue #288, opened 2024-11-22
Add support for dynamically defining the elements in the ProbeFields


JoaoBraveCoding:
## Request/Feature

Support for dynamically defining the  `ProbeFields` on a per placement or managed cluster basis.

## Use case

The observability group continues to develop the [MultiClusterObservabilityAddon (MCOA)](https://github.com/stolostron/multicluster-observability-addon), this addon has the ability to selectively deploy different observability signals collectors to different spoke clusters. Some observability deployment can be rather complex so we would want the ability to let users know if their deployments have been successful. To achieve this we started using the [AgentHealthProber](https://github.com/stolostron/multicluster-observability-addon/blob/main/internal/addon/addon.go#L52-L92) of type `HealthProberTypeWork`. Unfortunately this probe only supports a fixed list of `ProbeFields` which is not sufficient for our use case given some clusters might just want metrics while others might want metrics and logs.

zhiweiyin318:
This requirement is similar with https://github.com/open-cluster-management-io/addon-framework/issues/287
I have a PR https://github.com/open-cluster-management-io/addon-framework/pull/289 to address this.
In the PR, support wildcard in the name and namespace of ProbeField.ResourceIdentifier (depend on this https://github.com/open-cluster-management-io/ocm/issues/705), and add another func HealthChecker to WorkHealthProber which can check all feedback results at once.
@JoaoBraveCoding  could you help check if this can address your case?

JoaoBraveCoding:
@zhiweiyin318 thank you for the quick work! I've already [drafted a PR](https://github.com/stolostron/multicluster-observability-addon/pull/94) and tested this and it seems to do the job! I'll double check with the team to be sure 👍 Great work! 🙌

Issue #287, opened 2024-11-15
support different ProbeFields in WorkHealthProber


zhiweiyin318:
There is use case: 
I have an addon, the installStrategy is to deploy a deployment on some clusters, and deploy an OLM subscription on another clusters. 
So need to check the deployment status or the OLM subscription to do healthProber. 
Currently the FeedbackRule of the ProbeFields are `AND` relationship, but need "OR" in my case.

zhiweiyin318:
cc @tesshuflower feel free to append the details about the use case.

zhujian7:
/assign @zhiweiyin318 
/close
Fixed by https://github.com/open-cluster-management-io/addon-framework/pull/342

PR #289, opened 2024-11-23
:sparkles: add new addonHealthCheck func to check all fields and support wildcard

zhiweiyin318:
## Summary
1. add new addonHealthCheck func to check all filed results 
2. support wildcard in the ProbeField

## Related issue(s)
https://github.com/open-cluster-management-io/addon-framework/issues/288
https://github.com/open-cluster-management-io/addon-framework/issues/287
Fixes #

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
+++ b/pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
@@ -3,6 +3,7 @@ package agentdeploy
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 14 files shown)


pkg/agent/inteface.go
```diff
@@ -169,13 +169,19 @@ type HealthProber struct {
 }
 
 type AddonHealthCheckFunc func(workapiv1.ResourceIdentifier, workapiv1.StatusFeedbackResult) error
+type AddonHealthCheckerFunc func([]ResultField) error
```
zhujian7: Is the managedclusteraddon or managedcluster also required in some cases?
I remember someone raised a similar demand during our offline discussion?

pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
```diff
@@ -167,17 +168,38 @@ func (s *healthCheckSyncer) probeAddonStatusByWorks(
 		manifestConditions = append(manifestConditions, work.Status.ResourceStatus.Manifests...)
 	}
 
-	probeFields, healthChecker, err := s.analyzeWorkProber(s.agentAddon, cluster, addon)
+	probeFields, healthChecker, healthAllChecker, err := s.analyzeWorkProber(s.agentAddon, cluster, addon)
```
zhujian7: I don't think `healthAllChecker` is a proper name here. seems no meaning for `All`?

pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
```diff
@@ -167,17 +168,38 @@ func (s *healthCheckSyncer) probeAddonStatusByWorks(
...
 	if err != nil {
 		// should not happen, return
 		return err
 	}
 
+	var resultFields []agent.ResultField
+
 	for _, field := range probeFields {
```
zhujian7, line 180: can we separate the healthChecker(Deprecated one) and healthAllChecker logic into two different functions, so it is clear and easy to deprecate in the future?

pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
```diff
@@ -294,27 +331,46 @@ func (s *healthCheckSyncer) analyzeWorkloadsWorkProber(
...
 
-		if len(status.StatusFeedbacks.Values) == 0 {
-			return nil
-		}
+	return results
+}
 
-		return &status.StatusFeedbacks
+// compare two string, target may include *
+func wildcardMatch(resource, target string) bool {
```
zhujian7, line 357: These seem to be duplicates of the ocm/work code?

pkg/agent/inteface.go
zhiweiyin318: @qiujian16  looks like it makes sense to input cluster and mca. WDYT?

pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
zhiweiyin318: Add a TODO here, will follow up to refactor this part

pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
zhiweiyin318: yes. only this two parts use it

pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
zhiweiyin318: re-named

PR #297, opened 2025-01-09
✨ support addon agent resource requirements settings

elgnay:
## Related issue(s)

Fixes #


pkg/addonfactory/addondeploymentconfig.go
```diff
@@ -109,6 +119,47 @@ func ToAddOnProxyConfigValues(config addonapiv1alpha1.AddOnDeploymentConfig) (Va
...
+		return nil, nil
+	}
+
+	resourceRequirements, err := getRegexResourceRequirements(config.Spec.ResourceRequirements)
+	if err != nil {
+		return nil, err
+	}
+
+	type global struct {
+		ResourceRequirements []regexResourceRequirements `json:"resourceRequirements"`
```
qiujian16, line 148: does this need to be implemented for all addons?

pkg/addonfactory/addondeploymentconfig.go
```diff
@@ -22,6 +22,16 @@ var AddOnDeploymentConfigGVR = schema.GroupVersionResource{
 	Resource: "addondeploymentconfigs",
 }
 
+type resourceRequirements struct {
+	Limits   map[string]string `json:"limits,omitempty"`
+	Requests map[string]string `json:"requests,omitempty"`
+}
+
+type regexResourceRequirements struct {
+	ContainerIDRegex string               `json:"containerIDRegex"`
```
qiujian16: can we add some comment on what ContainerIDRegex means

pkg/addonfactory/addondeploymentconfig.go
elgnay: The answer is Yes for all helm chart based add-ons that want to support the resource requirements settings.

pkg/addonfactory/addondeploymentconfig.go
elgnay: Comments added.

pkg/addonfactory/addondeploymentconfig.go
qiujian16: @zhiweiyin318 what is the best place to doc this? it seems to me this is hidden in the code and hard for the user to understand.

zhiweiyin318: I think we can add it to the chapter `Values definition` here https://open-cluster-management.io/docs/developer-guides/addon/ .

PR #302, opened 2025-02-27
:warning: breaking change: when annotation is empty default managed by addon-manager

haoqing0110:
## Related issue(s)

Ref: https://github.com/open-cluster-management-io/ocm/issues/510


qiujian16:
kinda breaking changes, should title as breaking change I think.

haoqing0110:
/assign @qiujian16 
cc @zhujian7

PR #310, opened 2025-04-22
🐛 return err if get manifestconfigoption failed

zhujian7:
## Related issue(s)

Fixes #


pkg/addonmanager/controllers/agentdeploy/utils.go
```diff
@@ -491,7 +491,7 @@ func getManifestConfigOption(agentAddon agent.AgentAddon,
 
 		manifests, err := agentAddon.Manifests(cluster, addon)
 		if err != nil {
-			return manifestConfigs
+			return manifestConfigs, fmt.Errorf("get all deployments error: %v", err)
```
yanmxa, line 494: Will the GetValues error propagate here, right?

zhujian7: yes

PR #326, opened 2025-07-23
🌱 Upgrade golang to 1.24 and helm to 3.18.4

zhujian7:
## Related issue(s)

Fixes #


qiujian16:
I think we have to merge it given it is a security issue?

zhujian7:
> I think we have to merge it given it is a security issue?

yes, I think it is better to merge it!
/hold cancel

PR #332, opened 2025-08-27
:seedling: support grpc csr approval

skeeey:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

- New Features
  - Auto-approve certificate signing requests (CSRs) originating from the gRPC server service account when the requester is validated, while preserving existing approval behavior.

- Tests
  - Added test coverage for the new gRPC CSR approval path.

- Chores
  - Updated direct dependencies to newer versions.
  - Removed an unused indirect dependency.


pkg/utils/csr_helpers.go
```diff
@@ -173,10 +176,21 @@ func DefaultCSRApprover(agentName string) agent.CSRApproveFunc {
 		if strings.HasPrefix(csr.Spec.Username, "system:open-cluster-management:"+cluster.Name) {
 			klog.Info("CSR approved")
 			return true
-		} else {
-			klog.Info("CSR not approved due to illegal requester", "requester", csr.Spec.Username)
-			return false
 		}
+
```
qiujian16, line 182: can we do like

```go
userName := csr.Spec.Username
if csr.Spec.Username == defaultGRPCServiceAccount {
  userName = csr.Annotations[operatorapiv1.CSRUsernameAnnotation]
}

if strings.HasPrefix(userName, "system:open-cluster-management:"+cluster.Name) {
    return true
}
```

skeeey: updated

PR #333, opened 2025-09-10
⚠️ Update CSR interfaces to support enhanced context and error handling

zhujian7:
## Summary

  - Updates CSRSign interface to include cluster and addon context parameters with error return
  - Enhances CSRConfigurations interface to accept addon parameter and return errors
  - Improves error handling and messaging throughout CSR processing

  ### Why do we need this
  **The template-type addons need to read the CSR configurations and signing information from the AddonTemplate. Currently, it finds the template from the ClusterManagementAddon, which is not 100% accurate, as the addontemplate could be overridden in the ManagedClusterAddon config. So with this change, we can find the accurate AddonTemplate from the managedclusteraddon status.**

  Fix [this TODO](https://github.com/open-cluster-management-io/ocm/blob/b506d16cf8c6f908183d987a05af95727c9ce9ce/pkg/addon/templateagent/registration.go#L236) in ocm

  ### Changes
  #### CSRSign Interface
  - **Before:** `CSRSign(csr *CertificateSigningRequest) []byte`
  - **After:** `CSRSign(cluster *ManagedCluster, addon *ManagedClusterAddOn, csr *CertificateSigningRequest) ([]byte, error)`

  #### CSRConfigurations Interface
  - **Before:** `CSRConfigurations(cluster *ManagedCluster) []RegistrationConfig`
  - **After:** `CSRConfigurations(cluster *ManagedCluster, addon *ManagedClusterAddOn) ([]RegistrationConfig, error)`

  ### Migration Guide
  **CSRSign implementations:**
  1. Add cluster and addon parameters (can be ignored if not needed)
  2. Return ([]byte, error) instead of just []byte
  3. Handle errors properly in your signing logic

  **CSRConfigurations implementations:**
  1. Accept both cluster and addon parameters
  2. Return ([]RegistrationConfig, error) instead of just []RegistrationConfig
  3. Use addon parameter for addon-specific configuration logic when needed
  4. Return appropriate errors when configuration generation fails

## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/1179

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Certificate signing now receives cluster and addon context and reports errors for safer, context-aware issuance.
  * Registration configuration callbacks now accept addon context and return errors.
  * Agent install namespace callback now returns a value and error.

* **Refactor**
  * Standardized callbacks to return (result, error) for clearer error propagation across signing and registration flows.

* **Tests**
  * Unit and integration tests updated to the new callback shapes and to validate error handling.

### DIFF
--- a/pkg/agent/inteface.go
+++ b/pkg/agent/inteface.go
@@ -98,11 +98,15 @@ type AgentAddonOptions struct {
 	ConfigCheckEnabled bool
 }
 
+type CSRConfigurationsFunc func(cluster *clusterv1.ManagedCluster) ([]addonapiv1alpha1.RegistrationConfig, error)
+
 type CSRSignerFunc func(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn, csr *certificatesv1.CertificateSigningRequest) ([]byte, error)
 
 type CSRApproveFunc func(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn, csr *certificatesv1.CertificateSigningRequest) bool
 
 type PermissionConfigFunc func(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn) error
+
+type AgentInstallNamespaceFunc func(addon *addonapiv1alpha1.ManagedClusterAddOn) (string, error)
 
 // RegistrationOption defines how agent is registered to the hub cluster. It needs to define:
 // 1. csr with what subject/signer should be created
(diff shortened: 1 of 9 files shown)


pkg/agent/inteface.go
```diff
@@ -98,12 +98,19 @@ type AgentAddonOptions struct {
 	ConfigCheckEnabled bool
 }
 
-type CSRSignerFunc func(csr *certificatesv1.CertificateSigningRequest) []byte
+type CSRConfigurationsFunc func(cluster *clusterv1.ManagedCluster,
+	addon *addonapiv1alpha1.ManagedClusterAddOn) ([]addonapiv1alpha1.RegistrationConfig, error)
 
-type CSRApproveFunc func(cluster *clusterv1.ManagedCluster, addon *addonapiv1alpha1.ManagedClusterAddOn, csr *certificatesv1.CertificateSigningRequest) bool
+type CSRSignerFunc func(cluster *clusterv1.ManagedCluster,
```
qiujian16, line 104: this will change the signature of the function, how do we resolve compatibility problem

zhujian7: This is a breaking change; it is easy to migrate to the new version. The added cluster and addon parameters can be ignored if not needed. However, the developer still needs to make some slight changes.

qiujian16: let's ensure we add this to the release note.

PR #342, opened 2025-10-24
🐛 Fix health check to support partial probe results with multi-probe configuration

### LINKED ISSUE #341: cannot support empty FeedbackRule when there are multi  health check ProbeFields in 1.1.0
the use case: 
there are 2 probes to check deployment and OLM, but only one has feedback results in different cluster, another is empty. 

issue: 
the Available condition of addon is always unknown when one of the feedback result meet the health check. 

expect: 
the Available condition of addon is True.

In this case , [https://github.com/open-cluster-management-io/addon-framework/pull/324/files#diff-5588e2c7723c1aee20922dd8b5703[…]5f7076f577e1f7a5b6203f13c7614cR196](https://github.com/open-cluster-management-io/addon-framework/pull/324/files#diff-5588e2c7723c1aee20922dd8b5703522915f7076f577e1f7a5b6203f13c7614cR196)
If the feedback result for the first probe is empty, it will return immediately and set the status to something with "Probe results are not returned".
In the case, 1st probe may be empty, but the second probe might be the one I actually want. 
 
related issue:  https://github.com/open-cluster-management-io/addon-framework/issues/287

zhujian7:
## Summary

This PR fixes issue #341 where the health check incorrectly marks addons as Unknown when using multiple health check probes and the first probe returns empty feedback results.

**Problem:**
When multiple health check ProbeFields are configured (e.g., checking both deployment and OLM), if the first probe returns empty feedback results, the code immediately returns and sets the addon status to Unknown. This prevents the evaluation of subsequent probes that may have valid results.

**Solution:**
- Modified the logic to collect all empty probe fields instead of returning immediately
- Only mark the addon as Unknown if ALL probes have no results
- Allow partial probe results to be considered valid (if at least one probe returns results)
- Removed redundant check for non-empty results in healthChecker path

**Changes:**
1. Added `emptyProbeFields` slice to track probes with no results
2. Changed early return to `continue` when a probe has no results
3. Added check after the loop to determine if all probes are empty
4. Improved error messaging to differentiate between single vs multiple empty probes
5. Added comprehensive test case for the multi-probe scenario

## Related issue(s)

Fixes #341

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Health checks now aggregate results from multiple probes before deciding availability, avoiding immediate "unknown" reports and allowing partial probe success to mark an add-on available.
  * Consolidated empty-probe handling into a single post-check status update with a clearer message.

* **Tests**
  * Added a test validating multi-probe aggregation when initial probes are empty but later probes succeed.

* **Integration**
  * Updated integration expectation: unprobeable add-ons now treated as unavailable (False) instead of Unknown.

### DIFF
--- a/pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
+++ b/pkg/addonmanager/controllers/agentdeploy/healthcheck_sync.go
@@ -188,28 +188,20 @@ func (s *healthCheckSyncer) probeAddonStatusByWorks(
 	}
 
 	var FieldResults []agent.FieldResult
+	var emptyProbeFields []workapiv1.ResourceIdentifier
 
 	for _, field := range probeFields {
 		results := findResultsByIdentifier(field.ResourceIdentifier, manifestConditions)
 		// if no results are returned. it is possible that work agent has not returned the feedback value.
-		// mark condition to unknown
+		// collect these fields and check if all probes are empty later
 		if len(results) == 0 {
-			meta.SetStatusCondition(&addon.Status.Conditions, metav1.Condition{
-				Type:   addonapiv1alpha1.ManagedClusterAddOnConditionAvailable,
-				Status: metav1.ConditionUnknown,
-				Reason: addonapiv1alpha1.AddonAvailableReasonNoProbeResult,
-				Message: fmt.Sprintf("Probe results are not returned for %s/%s: %s/%s",
-					field.ResourceIdentifier.Group, field.ResourceIdentifier.Resource,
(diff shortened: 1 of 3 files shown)


zhiweiyin318:
feel free to help review this.  cc @tesshuflower

tesshuflower:
/lgtm

I've also done some tests in my env with this fix and it's working well - Thanks @zhujian7 and @zhiweiyin318 for such a quick fix!

qiujian16:
do we need to cherry-pick to release-1.1?

PR #381, opened 2026-05-01
:bug: addon fix for Install namespace potential race condition with addondeploymentconfig - For issue: https://github.com/open-cluster-management-io/ocm/issues/1465

tesshuflower:
<!--
Thanks for creating a pull request!

If this is your first time, please make sure to review CONTRIBUTING.MD.

Please copy the appropriate `:text:` or icon to the beginning of your PR title:

:sparkles: ✨ feature
:bug: 🐛 bug fix
:book: 📖 docs
:memo: 📝 proposal
:warning: ⚠️ breaking change
:rocket: 🚀 release
:seedling: 🌱 other/misc
:question: ❓ requires manual review/categorization

-->
## Summary

Fixes a race condition where addon controllers use the default namespace (open-cluster-management-agent-addon) before AddonDeploymentConfig has been synced to MCA status by the  addonconfiguration controller.                                                                                                                                                        
                                                                                                                                                                                        
When the registration controller (addon-framework) reconciles an MCA before the addonconfiguration controller (OCM addon-manager) has written configReferences, GetDesiredAddOnDeploymentConfig returns nil, nil — meaning "no config exists." Callers fall back to the default namespace, which is wrong if an AddonDeploymentConfig with a custom   
  namespace is actually configured.                                                                                                                                                     
                                                                                                                                                                                        
The fix checks two signals in MCA status before accepting an empty configReferences as authoritative:                                                                                 
  - supportedConfigs — does the addon declare support for addondeploymentconfigs?                                                                                                       
  - Configured condition — has the addonconfiguration controller set it to True?                                                                                                        
                                                                                                                                                                    
If the addon supports deployment configs but Configured is not yet True, an error is returned so callers requeue instead of proceeding with stale data. The check is in  GetDesiredAddOnDeploymentConfig to protect all three callers: AgentInstallNamespaceFromDeploymentConfigFunc, GetAddOnDeploymentConfigValues, and GetAgentImageValues. 


Things to note:

- This does return a new error in the case where we determine that the addonconfiguration controller has not yet updated. The code was put into GetDesiredAddOnDeploymentConfig() instead of AgentInstallNamespaceFromDeploymentConfigFunc directly.  this means it does affect more callers as indicated above.
- To determine if the configReferences have not been populated yet, we are using the .status.Condition `Configured` and checking that it exists and is True.  This means potentially we keep reconciling until this happens. Is there any potential issue here? It seems the configurationcontroller updates the status.configReferences and adds the Configured condition at the same time.
- Additionally, this requires the caller to retry on error - this does generally happen here: https://github.com/open-cluster-management-io/addon-framework/blob/bbaaa1ea338e24bc03e24ffec46605d867277d56/pkg/addonmanager/controllers/registration/controller.go#L278-L281.   However other callers could choose to ignore I suppose.

## Related issue(s)

Fixes #

https://github.com/open-cluster-management-io/ocm/issues/1465

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved addon deployment configuration handling: the app now distinguishes missing configuration from unsupported addons, triggers retries when an addon supports deployment config but isn't yet configured, and continues gracefully for unsupported or already-configured cases.

* **Tests**
  * Added comprehensive tests covering supported/unsupported addon scenarios, missing or false configured conditions, and various config-reference states to validate retry and defaulting behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/pkg/utils/addon_config.go
+++ b/pkg/utils/addon_config.go
@@ -49,15 +49,6 @@ func AgentInstallNamespaceFromDeploymentConfigFunc(
 		}
 
 		if config == nil {
-			// If the addon declares support for addondeploymentconfigs but the Configured condition
-			// is not yet True, the addonconfiguration controller either hasn't processed this MCA
-			// or hasn't finished rolling out configs. In either case configReferences may be
-			// incomplete. Return an error to trigger a requeue rather than falling back to the
-			// default namespace, which would be wrong if a config with a custom namespace exists.
-			if addonSupportsDeploymentConfig(addon) && !addonConfiguredTrue(addon) {
-				return "", fmt.Errorf("addon %s supports addondeploymentconfigs but configuration has not been synced yet, need to retry.",
-					addon.Name)
-			}
 			klog.V(4).InfoS("Addon deployment config is nil, return an empty string for agent install namespace",
 				"addonNamespace", addon.Namespace, "addonName", addon.Name)
 			return "", nil
@@ -76,6 +67,14 @@ func GetDesiredAddOnDeploymentConfig(
 	ok, configRef := GetAddOnConfigRef(addon.Status.ConfigReferences,
 		AddOnDeploymentConfigGVR.Group, AddOnDeploymentConfigGVR.Resource)
 	if !ok {
+		// If the addon declares support for addondeploymentconfigs but the Configured condition
+		// is not yet True, the addonconfiguration controller either hasn't processed this MCA
+		// or hasn't finished rolling out configs. In either case configReferences may be
+		// incomplete. Return an error so callers retry rather than proceeding with no config.
+		if addonSupportsDeploymentConfig(addon) && !addonConfiguredTrue(addon) {
+			return nil, fmt.Errorf("addon %s supports addondeploymentconfigs but Configured condition is not True yet, need to retry",
+				addon.Name)
+		}
 		return nil, nil
 	}
 
--- a/pkg/utils/addon_config.go
+++ b/pkg/utils/addon_config.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -47,12 +48,16 @@ func AgentInstallNamespaceFromDeploymentConfigFunc(
 			return "", fmt.Errorf("failed to get deployment config for addon %s: %v", addon.Name, err)
 		}
 
-		// For now, we have no way of knowing if the addon depleoyment config is not configured, or
-		// is configured but not yet been added to the managedclusteraddon status config references,
-		// we expect no error will be returned when the addon deployment config is not configured
-		// so we can use the default namespace.
-		// TODO: Find a way to distinguish between the above two cases
 		if config == nil {
+			// If the addon declares support for addondeploymentconfigs but the Configured condition
+			// is not yet True, the addonconfiguration controller either hasn't processed this MCA
+			// or hasn't finished rolling out configs. In either case configReferences may be
+			// incomplete. Return an error to trigger a requeue rather than falling back to the
+			// default namespace, which would be wrong if a config with a custom namespace exists.
+			if addonSupportsDeploymentConfig(addon) && !addonConfiguredTrue(addon) {
+				return "", fmt.Errorf("addon %s supports addondeploymentconfigs but configuration has not been synced yet, need to retry.",
+					addon.Name)
+			}
 			klog.V(4).InfoS("Addon deployment config is nil, return an empty string for agent install namespace",
 				"addonNamespace", addon.Namespace, "addonName", addon.Name)
 			return "", nil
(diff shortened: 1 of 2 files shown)


qiujian16 (commented):
/assign @haoqing0110

mikeshng:
@tesshuflower if the retrigger bot is not working, just let me know (here or DM if I missed the email). I can retry it. Hopefully when you are in the org, you will be able to do this.

tesshuflower:
> @tesshuflower if the retrigger bot is not working, just let me know (here or DM if I missed the email). I can retry it. Hopefully when you are in the org, you will be able to do this.

Thanks Mike! I've been trying to run the e2e locally to dig into it a bit - my change in this PR could cause the addon to take longer to reconcile as it needs to wait for the config controller to set the status - but I think it should work in general, even in this particular e2e test.

tesshuflower:
@qiujian16 I'm still looking for a review on this one, to verify that the approach of using the status conditions "Configured" set to True as the indicator that the config controller is done setting the configReferences - and therefore that we can trust the value there.

pkg/utils/addon_config.go
```diff
@@ -134,3 +138,17 @@ func GetAddOnConfigRef(
 
 	return false, addonapiv1beta1.ConfigReference{}
 }
+
+func addonSupportsDeploymentConfig(addon *addonapiv1beta1.ManagedClusterAddOn) bool {
+	for _, sc := range addon.Status.SupportedConfigs {
```
haoqing0110, line 143: addon.Status.SupportedConfigs is for addon users to know to supported config type. In code level, https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/addonmanager/controllers/addonconfig/controller.go#L177 , should check the configGVRs.

pkg/utils/addon_config.go
```diff
@@ -71,6 +67,14 @@ func GetDesiredAddOnDeploymentConfig(
 	ok, configRef := GetAddOnConfigRef(addon.Status.ConfigReferences,
 		AddOnDeploymentConfigGVR.Group, AddOnDeploymentConfigGVR.Resource)
 	if !ok {
+		// If the addon declares support for addondeploymentconfigs but the Configured condition
+		// is not yet True, the addonconfiguration controller either hasn't processed this MCA
+		// or hasn't finished rolling out configs. In either case configReferences may be
+		// incomplete. Return an error so callers retry rather than proceeding with no config.
+		if addonSupportsDeploymentConfig(addon) && !addonConfiguredTrue(addon) {
+			return nil, fmt.Errorf("addon %s supports addondeploymentconfigs but Configured condition is not True yet, need to retry",
```
qiujian16, line 75: can we have an integration test on this? Also if it returns err, the controller will backoff upon error. Instead of returning an error, should we return a state that "nothing is configured yet" so the caller will know nothing should be handled, and when addon status is updated, this func will be triggered again.

pkg/utils/addon_config.go
tesshuflower: Thanks for this @haoqing0110  - I guess I was having trouble seeing why the configGVRs would get loaded in this code - but using Status.SupportedConfigs has a similar timing issue as it's something else that's setting that status.

I think in the end, are you ok if we use `configured=True` condition to decide that the configReferences are ready to be consumed?  It seems that this is the correct condition to check.

However, there is the one caveat that where a ClusterManagementAddOn may have this annotation: `addon.open-cluster-management.io/lifecycle: "self"` - in this case, the external controller does not ever set `configured=True` condition.

Do you think this is a concern? I think with v1beta1 this is less of a concern since there is no such idea of "supportedConfigs" anymore - and if the annotation is set by a user, do we not expect them to call `AgentInstallNamespaceFromDeploymentConfigFunc()` at all?

pkg/utils/addon_config.go
tesshuflower: Yes, this is a concern I mentioned in my description as well - the issue is the function `AgentInstallNamespaceFromDeploymentConfigFunc()` is essentially a helper function provided by the addon-framework that addons are already using to get the namespace from the deploymentconfig - Should I consider deprecating it and making a new function instead that can return the extra status?

pkg/utils/addon_config.go
haoqing0110: @tesshuflower I think it's good to check the condition `configured=True`. And after migrating to v1beta1, the code won't have the annotation any more. https://github.com/open-cluster-management-io/ocm/issues/1428

pkg/utils/addon_config.go
tesshuflower: @qiujian16 What do you think about my comment above?  My main concern with changing the function signature is this is a function already called by users of the addon-framework - If I want to return another parameter to tell them to retry, this necessitates a change in the function or perhaps a new one.  Do you have a preference here?

qiujian16: yes I think it makes sense.

tesshuflower: @qiujian16 @haoqing0110 I've come back to this one, and was going through changes required to add new functions to be able to get the namespace (with retries) - however there are a lot of cascading affected places.

For example addon framework interface functions would need updating too, example: https://github.com/open-cluster-management-io/addon-framework/blob/929daf05d87b325d61bc1f291c429144f1e2e202/pkg/agent/interface.go#L73

And then even from getValues functions like here: https://github.com/open-cluster-management-io/addon-framework/blob/929daf05d87b325d61bc1f291c429144f1e2e202/pkg/addonfactory/helm_agentaddon.go#L170
Essentially all of those don't currently have a strict retry mechanism and would need to change or get updated.

But then I found we actually have this `ConfigCheckEnabled` setting already:

https://github.com/open-cluster-management-io/addon-framework/blob/929daf05d87b325d61bc1f291c429144f1e2e202/pkg/agent/interface.go#L112

And this is used in a couple places before calling the agent Manifests() function.

Do you think perhaps I should pivot to a fix that simply uses ConfigCheckEnabled instead? Ultimately it's the same exact guard - it checks for configured=true in the status before proceeding.  I feel like my current fix in this PR is maybe bypassing the `ConfigCheckEnabled as well as getting overly complicated.

Essentially a fix would be to make sure we do something similar to this:

https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/addonmanager/controllers/agentdeploy/controller.go#L409-L413

Anytime before the getAgentNamespace (of even agent Manifests()) is called.  This would essentially give us the same protection, but would only be turned on when users set `ConfigCheckEnabled`.


Sorry the above is very long, I've made an alternative draft PR here to demonstrate the changes required with this approach: https://github.com/open-cluster-management-io/addon-framework/pull/382

The comments on the original function with the race condition attempt to explain it as well: https://github.com/open-cluster-management-io/addon-framework/pull/382/changes#diff-0268a17aba5249e6bd421172ca6ac3c9c0e93ef819030b1cba41afbe50d89743

tesshuflower: @qiujian16 and @haoqing0110 - sorry do you mind taking a look again at the above? Would you be ok with the alternative approach in #382 ?

PR #385, opened 2026-06-12
:seedling: chore: upgrade to go 1.26

tesshuflower:
Upgrades Go version from 1.25 to 1.26.

Changes:
- `go.mod` — go directive updated to 1.26.0
- `.github/workflows/go-presubmit.yml` — GO_VERSION updated to 1.26
- `.github/workflows/go-postsubmit.yml` — GO_VERSION updated to 1.26
- `.github/workflows/go-release.yml` — GO_VERSION updated to 1.26
- `.github/workflows/cloudevents-integration.yml` — GO_VERSION updated to 1.26 (was incorrectly pinned to 1.24)
- `build/Dockerfile.example` — base image updated to golang:1.26-bookworm

Relates to: https://github.com/open-cluster-management-io/ocm/issues/1555

## Summary by CodeRabbit

* **Chores**
  * Updated Go toolchain version to 1.26 across build configurations and CI/CD workflows.

PR #386, opened 2026-06-12
:seedling: chore: pin GitHub Actions to digest SHAs and replace setup-kind

tesshuflower:
Pin all GitHub Actions to immutable commit SHAs to prevent supply chain issues from moving tags.

Replace `engineerd/setup-kind` with a direct `kind create cluster` run step using the `kind` binary pre-installed on GitHub-hosted runners. The Kubernetes node image version is now explicit via `KUBERNETES_VERSION` env variable set to `v1.35.0` (same version previously used by kind v0.31.0 by default).

Removing setup-kind GH action as it appears to no longer be maintained, and now is causing issues when we use v0.6.2 as the tag v0.6.2 points to a version that is not correct (actual version is the branch v0.6.2). Note both are old and last modified in 2024.

~~> **Note:** Draft PR — to be merged after the Go 1.26 upgrade PR (#385) merges to avoid conflicts.~~

### DIFF
--- a/.github/workflows/cloudevents-integration.yml
+++ b/.github/workflows/cloudevents-integration.yml
@@ -23,9 +23,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: install Go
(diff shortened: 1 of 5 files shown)


tesshuflower:
@qiujian16 or @haoqing0110 could I get a review on this one? Thanks!

PR #388, opened 2026-06-26
:bug: Fix GetAddOnConfigRef selection order

### LINKED ISSUE #387: GetAddOnConfigRef returns first matching ConfigReference instead of last
**Describe the bug**

`GetAddOnConfigRef` returns the **first** matching `ConfigReference` for
a given group+resource, but the documented contract on
`GetAddOnDeploymentConfigValues` states that the **last** entry should
take precedence.

The comment on `GetAddOnDeploymentConfigValues` (in
`pkg/addonfactory/addondeploymentconfig.go`) says:

> *"If there are multiple AddOnDeploymentConfig objects in the AddOn
> ConfigReferences, the big index object will override the one from small
> index"*

However, `GetAddOnConfigRef` (in `pkg/utils/addon_config.go`) iterates
forward and returns on first match, so the *small* index object always
wins. The function's own TODO comment acknowledges this:

> *"(TODO) this needs to be reconsidered if we support multiple same GK
> in the config references"*

Multiple same-GK config references have been supported since OCM v0.15.0,
making this a functional bug.

**To Reproduce**

Prerequisites: a hub cluster with OCM addon-manager running. This example
uses `cluster1` as the managed cluster name, consistent with the
[OCM contributing guidelines](https://github.com/open-cluster-management-io/ocm/blob/main/CONTRIBUTING.md)
for kind-based test environments.

1. Create an `AddOnTemplate` with a variable placeholder:

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: AddOnTemplate
metadata:
  name: ordering-demo
spec:
  addonName: ordering-demo
  agentSpec:
    workload:
      manifests:
        - apiVersion: v1
          kind: ConfigMap
          metadata:
            name: ordering-demo
            namespace: open-cluster-management-agent-addon
          data:
            greeting: "{{GREETING}}"
EOF
```

2. Create two `AddOnDeploymentConfig` resources with different values:

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: AddOnDeploymentConfig
metadata:
  name: greeting-hello
  namespace: cluster1
spec:
  customizedVariables:
    - name: GREETING
      value: "hello"
---
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: AddOnDeploymentConfig
metadata:
  name: greeting-goodbye
  namespace: cluster1
spec:
  customizedVariables:
    - name: GREETING
      value: "goodbye"
EOF
```

3. Create a `ClusterManagementAddOn` referencing the template (no global
   ADC default, so per-cluster configs are the only source):

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ClusterManagementAddOn
metadata:
  name: ordering-demo
  annotations:
    addon.open-cluster-management.io/lifecycle: "addon-manager"
spec:
  addOnMeta:
    displayName: Config Ordering Demo
  installStrategy:
    type: Manual
  supportedConfigs:
    - group: addon.open-cluster-management.io
      resource: addontemplates
      defaultConfig:
        name: ordering-demo
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
EOF
```

4. Create the `ManagedClusterAddOn` with two ADC references — `hello`
   first, `goodbye` second:

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ManagedClusterAddOn
metadata:
  name: ordering-demo
  namespace: cluster1
spec:
  configs:
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
      name: greeting-hello
      namespace: cluster1
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
      name: greeting-goodbye
      namespace: cluster1
EOF
```

5. Wait for the addon to reconcile, then check which value was selected:

```bash
kubectl get configmap ordering-demo -n open-cluster-management-agent-addon -o jsonpath='{.data.greeting}'
```

The result is `hello` (the first entry), not `goodbye` (the last entry).

Per the documented contract, the last entry (`goodbye`) should take
precedence.

Cleanup:

```bash
kubectl delete clustermanagementaddons ordering-demo
kubectl delete addontemplates ordering-demo
kubectl delete addondeploymentconfigs greeting-hello greeting-goodbye -n cluster1
```

**Expected behavior**

When multiple config references of the same group+resource exist in
`status.configReferences`, the **last** entry should take precedence,
matching the documented contract on `GetAddOnDeploymentConfigValues`.

**Environment ie: OCM version, Kubernetes version and provider:**

- addon-framework (all versions with multi-GK config support)
- Kubernetes 4.x / OpenShift 4.x

**Additional context**

The root cause is in `pkg/utils/addon_config.go`:

```go
func GetAddOnConfigRef(
    configReferences []addonapiv1beta1.ConfigReference,
    group, resource string) (bool, addonapiv1beta1.ConfigReference) {

    for _, config := range configReferences {
        if config.Group == group && config.Resource == resource {
            return true, config    // returns on FIRST match
        }
    }
    return false, addonapiv1beta1.ConfigReference{}
}
```

The fix is to iterate the full list and return the last match:

```go
func GetAddOnConfigRef(
    configReferences []addonapiv1beta1.ConfigReference,
    group, resource string) (bool, addonapiv1beta1.ConfigReference) {

    found := false
    var result addonapiv1beta1.ConfigReference
    for _, config := range configReferences {
        if config.Group == group && config.Resource == resource {
            found = true
            result = config
        }
    }
    return found, result
}
```

**Related issues**

There is a separate bug in `ocm` where re-ordering
entries in `ManagedClusterAddOn.spec.configs` does not update the order
in `status.configReferences`, compounding this issue. See https://github.com/open-cluster-management-io/ocm/issues/1600

**Risks**
This PR does introduce a change in behaviour, however, I think it is unlikely that users are dependent on this selection order because the presence of a 2nd config currently has no impact (When the config is a AddOnDeploymentConfig, it would be nice if variables are merged between multiple configs - I opened an enhancement to implement that: https://github.com/open-cluster-management-io/ocm/issues/1603)

KevinFCormier:
## Summary
Corrects precedence of selection when there are multiple config references of the same group-kind. (Should be last in list instead of first.)

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/addon-framework/issues/387

## Summary by CodeRabbit

* **Bug Fixes**
  * When multiple matching config references exist, the app now uses the latest matching entry instead of the first one.
  * Improved handling when no matching reference is found, returning a clear “not found” result.

* **Tests**
  * Added coverage for single, missing, mixed, and duplicate match scenarios to verify the updated selection behavior.

### DIFF
--- a/pkg/utils/addon_config.go
+++ b/pkg/utils/addon_config.go
@@ -120,18 +120,22 @@ func GetAddOnDeploymentConfigSpecHash(config *addonapiv1beta1.AddOnDeploymentCon
 	})
 }
 
-// GetAddOnConfigRef returns the first addon config ref for the given config type. It is fine since the status will only
-// have one config for each GK.
-// (TODO) this needs to be reconcidered if we support multiple same GK in the config referencese.
+// GetAddOnConfigRef returns the last addon config ref for the given config type.
+// When multiple config references share the same group+resource, the highest-index
+// entry takes precedence, consistent with the contract documented on
+// GetAddOnDeploymentConfigValues.
 func GetAddOnConfigRef(
 	configReferences []addonapiv1beta1.ConfigReference,
 	group, resource string) (bool, addonapiv1beta1.ConfigReference) {
 
+	found := false
+	var result addonapiv1beta1.ConfigReference
 	for _, config := range configReferences {
 		if config.Group == group && config.Resource == resource {
-			return true, config
+			found = true
+			result = config
 		}
 	}
 
-	return false, addonapiv1beta1.ConfigReference{}
+	return found, result
 }
--- a/pkg/utils/addon_config_test.go
+++ b/pkg/utils/addon_config_test.go
@@ -23,6 +23,136 @@ func newTestAddOnDeploymentConfigGetter(adc *addonapiv1beta1.AddOnDeploymentConf
 	return &testadcGetter{adc: adc}
 }
 
+func TestGetAddOnConfigRef(t *testing.T) {
+	cases := []struct {
+		name             string
+		configReferences []addonapiv1beta1.ConfigReference
+		group            string
+		resource         string
+		expectedFound    bool
+		expectedName     string
+	}{
+		{
+			name:             "no config references",
+			configReferences: []addonapiv1beta1.ConfigReference{},
+			group:            "addon.open-cluster-management.io",
+			resource:         "addondeploymentconfigs",
+			expectedFound:    false,
+			expectedName:     "",
+		},
+		{
+			name: "no matching group+resource",
+			configReferences: []addonapiv1beta1.ConfigReference{
+				{
+					ConfigGroupResource: addonapiv1beta1.ConfigGroupResource{
+						Group: "addon.open-cluster-management.io", Resource: "addontemplates",
+					},
+					DesiredConfig: &addonapiv1beta1.ConfigSpecHash{
+						ConfigReferent: addonapiv1beta1.ConfigReferent{Name: "my-template"},
+						SpecHash:       "hash1",
+					},
+				},
+			},
+			group:         "addon.open-cluster-management.io",
+			resource:      "addondeploymentconfigs",
+			expectedFound: false,
+			expectedName:  "",
+		},
+		{
+			name: "single match",
+			configReferences: []addonapiv1beta1.ConfigReference{
+				{
+					ConfigGroupResource: addonapiv1beta1.ConfigGroupResource{
+						Group: "addon.open-cluster-management.io", Resource: "addondeploymentconfigs",
+					},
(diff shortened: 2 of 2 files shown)


tesshuflower:
@haoqing0110 or @qiujian16  - this change looks good to me, but could one of you verify?  Just want to confirm that this is the intended ordering (rather than the docs being the place that is incorrect).

I have the same question when it comes to the ClusterManagementAddOn:

`spec.installStrategy.placements[].configs` <- what's the expected behaviour here if there are multiple configs with the same GVK?

PR #389, opened 2026-07-21
:bug: normalize Helm values before schema validation

kahirokunn:
## Summary

Helm schema validation rejects concrete Go collections returned by `GetValuesFuncs` with `invalid jsonType`, forcing each add-on to merge and normalize values itself; normalize the merged values in addon-framework instead.

## Related issue(s)

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Helm value handling by normalizing user-provided values into JSON-compatible forms before merging, fixing rendering for settings such as pod tolerations.
  * Better failure behavior and clearer error messages when Helm values contain unsupported data types.
* **Documentation**
  * Clarified that returned Helm values may be typed and are normalized to JSON-compatible types prior to validation/rendering.
* **Tests**
  * Expanded coverage for collection-based values, tolerations, and normalization error scenarios (including expected failures).
* **Chores**
  * Added a JSON Schema used to exercise Helm values validation in tests.

### DIFF
--- a/pkg/addonfactory/testmanifests/chart/values.schema.json
+++ b/pkg/addonfactory/testmanifests/chart/values.schema.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "$comment": "exercises helm schema validation; typed Go values that are not normalized to JSON-compatible types fail this validation",
+  "type": "object",
+  "properties": {
+    "tolerations": {
+      "type": "array"
+    }
+  }
+}
--- a/docs/helmAgentAddon.md
+++ b/docs/helmAgentAddon.md
@@ -45,6 +45,8 @@ Helm Chart built-in values
 
 In the list of `GetValuesFuncs`, the values from the big index Func will override the one from low index Func.
 The built-in values will override the values got from the list of `GetValuesFuncs`.
+The values returned by `GetValuesFuncs` may contain typed Go values such as `map[string]string` or structs with json tags; they are normalized to JSON-compatible types before merging.
+Values that cannot be marshaled to JSON will make the manifest rendering fail.
 
 The Variable names in Values should begin with lowercase. So the best practice is to define a json struct for the values, and convert it to Values using the `JsonStructToValues`.
 
@@ -52,4 +54,3 @@ The Variable names in Values should begin with lowercase. So the best practice i
 We support a helper `GetValuesFunc` named `GetValuesFromAddonAnnotation` which can get values from annotation of ManagedClusterAddon.
 The key of the Helm Chart values in annotation is `addon.open-cluster-management.io/values`,
 and the value should be a valid json string which has key-value format.
-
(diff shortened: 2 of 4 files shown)


kahirokunn:
@mikeshng Hi ✋ `value.schema.json` is an extremely powerful Helm feature, and I'd like to make it more user-friendly. If you're available, I'd be grateful for your feedback. Thank you 🙏

kahirokunn:
Thank you @coderabbitai . Fixed it.

pkg/addonfactory/testmanifests/chart/values.schema.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "$comment": "exercises helm schema validation; typed Go values that are not normalized to JSON-compatible types fail this validation",
```
mikeshng, line 3: I welcome this enhancement.

My biggest concern will be any backward compatibility issues for existing addons? Any migration step require? I am worry that if we upgrade to this addon-framework version with this change and it might break existing addons unless they do something. 

If it's possible to break existing addons, could we enhance the PR so that we get this feature and also not break existing users. That will be ideal.

/assign @qiujian16

kahirokunn: Thanks for raising this.

Existing add-ons that return JSON-compatible values will keep working without any migration. This includes annotation values and values already converted with `JsonStructToValues`. `values.schema.json` also remains optional; this change does not enable schema validation for existing charts.

I also checked `addon-contrib`. There is currently no tracked use of `BuildHelmAgentAddon`; its `addonfactory` consumers all use `BuildTemplateAgentAddon`, so they are not affected.

For schema-enabled Helm add-ons, this removes the need for local normalization. For example, cluster-proxy currently has [its own merge-and-normalize workaround](https://github.com/open-cluster-management-io/cluster-proxy/blob/1949eeffc71d8e768831d56948e2f6fa80f0fad9/pkg/proxyagent/agent/agent.go#L130-L163), although that change is still part of [cluster-proxy PR #329](https://github.com/open-cluster-management-io/cluster-proxy/pull/329).

Normalization is applied to Helm charts without a schema as well. In theory, values that cannot be marshaled to JSON will now fail, but those are outside the documented and recommended usage of Helm values. Concrete nested maps will also be deep-merged as intended by `MergeValues`.

Therefore, I do not expect this change to affect existing add-ons, and no migration should be required.

kahirokunn:
@coderabbitai review

pkg/addonfactory/testmanifests/chart/values.schema.json
mikeshng: /lgtm

Thank you for taking a look at my feedback.

I am not a maintainer for this repo so @qiujian16 will have to approve. Thanks.

qiujian16 (commented):
/approve
/lgtm

kahirokunn:
Thanks for your review 🙏

PR #398, opened 2026-09-19
:bug: fix: delete and recreate role bindings on RoleRef change

### LINKED ISSUE #397: ApplyRoleBinding and ApplyClusterRoleBinding fail to update on RoleRef change
pkg/utils/permission.go's ApplyRoleBinding and ApplyClusterRoleBinding both carry the comment "TODO on non-matching roleref, delete and recreate", but when roleRefIsSame is false, they just set existingCopy.RoleRef and call Update anyway.

RoleRef on a RoleBinding or ClusterRoleBinding is immutable in the Kubernetes RBAC API, so the apiserver rejects an Update that changes it. This means whenever an addon's required RBAC config changes which Role or ClusterRole a binding should point at, ApplyRoleBinding/ApplyClusterRoleBinding will fail with a validation error instead of reconciling correctly, which is exactly the case the existing TODO comment says still needs handling.

ApplyRole and ApplyClusterRole are unaffected, they only ever mutate Rules, which is mutable.

Neither function has any existing test coverage.

Fix should follow the TODO: when RoleRef differs, delete the existing binding and create the required one instead of calling Update.

magic-peach:
## Summary
ApplyClusterRoleBinding and ApplyRoleBinding call Update when the required RoleRef differs from the existing one, but RoleRef is immutable on RBAC bindings so the apiserver rejects the update. Delete and recreate the binding when RoleRef changes, matching the existing TODO comment.

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/addon-framework/issues/397

## Summary by CodeRabbit

* **Bug Fixes**
  * Role bindings are now recreated when their referenced role changes, ensuring the updated permissions take effect.
  * Changes to binding subjects continue to update the existing binding without replacing it.


mikeshng (commented):
/hold

Note to maintainers: do not remove hold until we discuss this further. Thanks!

magic-peach:
@mikeshng are any changes needed or something?

mikeshng (commented):
> @mikeshng are any changes needed or something?

No changes needed yet but we are usually more careful with these type of "hidden" behavior changes. Sorry I wasn't more specific in my last review.

magic-peach:
@mikeshng Thanks, do let me know what changes you would like from my end maybe we could emit a log line whenever a binding is deleted and recreated due to a RoleRef change, so it's visible in practice rather than a silent delete/create.
Or maybe add a note to the relevant godoc / addon developer docs explaining that changing the referenced Role/ClusterRole on an existing binding causes a delete+recreate

mikeshng (commented):
I have to think more about this one. Deleting/creating rolebindings even the higher privilege clusterrolebindings is not a trivial change. There is a reason why we put a TODO there. 

I don't see what's the gain of doing this vs the risk. Except for in the code there is a `TODO` so let's do it. That's not how we usually do things.

PR #399, opened 2026-09-20
:seedling: test: add coverage for KubeClient and CustomSigner registration config mapping

magic-peach:
## Summary

pkg/agent/registration.go's KubeClientRegistration.RegistrationAPI() and CustomSignerRegistration.RegistrationAPI() are pure struct to API converters used by every addon that declares its registration intent, but had zero test coverage. Added registration_test.go covering KubeClientRegistration with an empty user and groups, a user only, groups only, and both set, and CustomSignerRegistration with all fields set and with the optional fields omitted.

## Related issue(s)

No related issue, found while reviewing pkg/agent for test coverage gaps.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for agent registration API behavior.
  - Verified Kubernetes client registrations with user and group details, including empty-field handling.
  - Verified custom signer registrations with signer, user, group, and organizational unit details.
  - Added coverage for custom signer registrations containing only a signer name.


mikeshng (approved):
/approve

/lgtm

Sure. Have tests is better than not having  them.

